PR #18, opened 2026-08-19
One predicate per decision: data-layer surface, panel_roles, telemetry, dispatch routing (#17)

### LINKED ISSUE #17: One predicate per decision: data-layer surface, panel_roles, telemetry, dispatch routing
**Risk tier: architectural.** Forced by this repo's own `pipeline.config.json`
`architecturalTriggers.keywords` — this change touches `gate`, `schema`, `agent contract`,
`risk tier`, `panel composition` and `hook`-adjacent dispatch. Not a judgment call.

## Problem

Four decisions the orchestrator currently re-derives by hand at N sites. Each moves into one
deterministic, config-driven, testable predicate. An external teardown named the ratio of
deterministic enforcement points (~5) to imperative markdown rules (several dozen) as this
design's central structural weakness; this narrows that ratio.

### 1. The data-layer surface has three disagreeing definitions (the live defect)

Verified by execution at HEAD `75420f0`:

| path | tripwire `^migrations/` (`pipeline.md:556`) | panel `^(migrations/\|db/)` (`:597`,`:722`) | gate `**/migrations/**` (`gate-pre-phase4.mjs:65`) |
|---|---|---|---|
| `migrations/0001.sql` | MATCH | MATCH | MATCH |
| `db/migrations/0042.sql` | **no** | MATCH | MATCH |
| `supabase/migrations/x.sql` | **no** | **no** | MATCH |

- A `db/migrations/**` path seats DBA on the panel but **does not trip the mis-tier halt**, so
  a migration ships the standard lane with no Phase 2 fan-out and no *pre-code* DBA review.
- A `supabase/migrations/**` path is worse: **DBA is never seated at all**, at any point.

The tripwire's own preamble (`pipeline.md:549`) says a migration in a standard diff means "the
never-skip DBA migration gate was bypassed". It is right, and it does not fire. Its own
CUSTOMIZE comment at `:554` names `'^db/migrations/'` as the example — the plugin documents
the exact path shape its default fails on.

`pipeline.md:619` already states the governing principle, for frontend only: use "the SAME
allowlist the gate uses, so detection and dispatch never diverge." `scripts/frontend-surface.mjs`
is the fix for this exact class, already shipped. DBA and DevOps never got it.

Two dead config-key references compound it: `README.md:64` and `pipeline.md:928` both cite
`migrationsGlob`, which **does not exist** (code reads `migrationGlobs`). The README row is
wrong three ways — wrong name, wrong default (`migrations/**` vs `**/migrations/**`), and it
claims the key drives the mis-tier tripwire, which no config key does.

### 2. Two verified defects

- **Vacuous shard fallback.** `pipeline.md:647` tells a reviewer whose write is refused to fall
  back to `<WORKTREE_PATH>/.pipeline/<issue>/` — the identical directory `:409` already set as
  `ARTIFACT_DIR`. Since `merge-peer-review.mjs` exits 2 on a missing shard, the documented
  escape hatch produces a hard halt with a binding verdict stranded on disk.
- **`panel_roles` schema gap.** `schemas/status.schema.json:27` enumerates six roles;
  `pipeline.md` appends `design_review` (`:626`) and `art_director` (`:614`,`:630`). The schema
  does not merely omit them — its own description asserts "The design_review lens is tracked via
  its own shard", contradicting what the command actually writes. Latent (nothing validates
  status.json), but `knowledge-store.mjs:11` archives it verbatim, so schema-illegal values are
  already accumulating in the permanent knowledge store.

### 3. Self-telemetry (narrowly scoped)

`events[]` with per-phase ISO timestamps already exists AND already survives archival verbatim
(`knowledge-store.mjs:11` includes `status` in `ARCHIVE_ARTIFACTS`). The retention path works.
The gap is only that **nothing computes** per-phase elapsed or a round count from it.

### 4. Dispatch model pins live at seven sites

`pipeline.md:118`, `:359`, `:360`, `:669`, `:673` (five inline `model:` literals), `:678` (prose
restating all of them), and `README.md:70` (an independent seventh restatement), plus nine agent
frontmatter files they must stay consistent with. A routing table that does not rewire all seven
becomes an **eighth** source of truth instead of replacing seven — the same bug class this change
exists to fix, one layer up.

## Resolutions

**Q1 — no alias. Two independent keys.** `migrationGlobs` stays NARROW with its match set,
default and AC19 semantics untouched; it gains consumers (tripwire, voice reversibility) but not
a wider match set. `dataLayerGlobs` is a NEW broad key feeding panel composition only. Its
default is the union of the adopter's effective `migrationGlobs` and the broad data-layer
defaults, so an adopter who set `migrationGlobs:["db/changes/**"]` gets DBA seated for their
migrations with no new config. `migrationsGlob` stays a **misspelling** that config-doctor must
keep flagging — `tests/test-config-doctor.sh:57` asserts it suggests `migrationGlobs`, and
legitimizing the typo as an alias would defeat that shipped test.

**Q2 — confirmed, in scope.** Every new key registers in `config-doctor.mjs` `CODE_KEYS` with
`reader`/`fallback`/`degrades`, or the tool's own warmup reports it as "read by nothing" — the
exact failure config-doctor exists to catch, self-inflicted.

**Q3(a) — IN.** `agents/dba.md:33` ("the query layer") and `agents/devops.md:32` ("service/deploy
config") already claim a broader surface than the greps match. Verified: `devops.md`'s own worked
example at `:134`, `services/api/deploy.config`, does not match its own trigger regex. These are
the **requirements source** for the broad predicate, not passive readers.

**Q3(b) — code fix OUT, filed as #16.** The executable-down-section gate gap claimed as "tracked
as follow-up issue 4" was never filed; issue #4 is a merged voice fix. Now filed. The companion
false pointer ("follow-up doc issue 3" at `test-gate-pre-phase4.sh:317-318`, where issue #3 is
the autorun hook) covers a doc-only bug — `gate-pre-phase4.mjs:16-19` claims `migrationGlobs: []`
"disables it entirely", which is flatly wrong — and folds in here as a third wrong statement
about `migrationGlobs` beside the two already being corrected. Both false pointers get corrected
to real numbers.

## Behavior change, stated deliberately

The tripwire's match set **widens** (hard-coded `^migrations/` to the `migrationGlobs` default
`**/migrations/**`). It halts strictly more often, never less. That is the fix, and it is in the
fail-closed direction. Everything else ships behavior-neutral, including the routing table, which
encodes today's exact assignments; Haiku is unused on day one.

## Scope note

The four changes land as separately revertable commits, surface module first. The highest-severity
item (1, a live safety hole) is otherwise coupled to the highest-transcription-risk item (4), and
a Phase 4 `REQUEST_CHANGES` on the routing table must not block the migration-tripwire fix.

Full spec: `.pipeline/<issue>/spec.json`.

nsmedia-io:
Closes #17.

Four decisions the orchestrator re-derived by judgment at N sites each now have one predicate apiece.

## What changed

**One data-layer surface module** (`scripts/data-layer-surface.mjs`), parallel to the shipped `frontend-surface.mjs`, with **three named glob resolvers** whose names carry the difference a later refactor would otherwise collapse:

- `migrationGlobsForTripwire(cfg)` **UNIONS** config with a fifteen-row framework-preset set. Config can only ever WIDEN the halt. Binding a halting control to a narrowing knob is how a four-character edit disarms it while the config still reports healthy.
- `migrationGlobsForGate(cfg)` keeps the shipped **REPLACE** semantics the down-section gate is tested on.
- `dataLayerGlobs(cfg)` is the broad panel predicate, defaulting over the tripwire's union so an adopter's `migrationGlobs` widens the DBA seat automatically.

`extraMigrationGlobs` is a new additive key that unions into all three and can never replace.

**The rewired tripwire fails CLOSED.** It captures the module's stdout and exit status separately and branches on the status first. Written as `<node cmd> | grep -q`, a pipe would discard the exit status: an absent or throwing module exits non-zero with empty stdout, the condition reads false, and no halt fires, silently restoring the pre-fix state. An unevaluable module now halts with `current_phase: "3-impl-tripwire-indeterminate"`.

**The model resolver fails OPEN to frontmatter**, which is the opposite direction on purpose. `scripts/dispatch-model.mjs` replaces five inline model literals, a prose restatement and a README assignment list. `secops` and `qa` are pinned in code, emit no `model:` key at all, and are unreachable from config.

Plus: config-doctor reports a configured glob pattern that owns nothing (per pattern, per consumer, because a per-key check could never fire); `status.json` gains `review_rounds`, derived telemetry, and an `effective_config` audit record with the two migration sets as separate entries; the Phase 4 shard fallback now points somewhere different from `ARTIFACT_DIR` and the merge step reads it; and a CI workflow finally runs the test suite.

## The AC18 collision, resolved

The Phase 2.5 design sketches (sonnet) and the bake-off judge (opus) are both `subagent_type: "dev"` in phase 2.5, so a resolver keyed on `(role, risk_tier, phase)` cannot reproduce both. The key carries a fourth dimension, an optional `--site` label. Neither site's model was changed, because behaviour-neutrality is the property the table exists to preserve. A call that omits `--site` on a multi-site key resolves to the default row and REPORTS the ambiguity naming every site at that key.

## Tests

23 suites, 996 assertions passing. The 16 pre-existing suites are at their exact 531/0 baseline, per suite.

Two QA-authored assertions are red and are **unsatisfiable as written**; both are documented with evidence and a one-line fix in `.pipeline/17/impl-report.json` under `qa_test_defects`. One concatenates two `emission()` calls and expects a single token (that helper prints a token in all six of its reachable cells). The other's fixture removes the surface module but keeps the resolver, then asserts the behaviour of an absent resolver. Neither was edited.

## Behaviour changes for adopters

Three, all documented in README's new Upgrading section: the preset defaults widened and the tripwire can no longer be narrowed, `pipeline.config.json` is now an architectural trigger, and a copied example config's `migrationGlobs` line must be deleted.


nsmedia-io:
## Merge record

Verdict **APPROVE_WITH_NOTES** after 7 review rounds. SecOps lifted its VETO at `9d076c0`; the notes from that round were fixed in the nit round ending `4b721b7`.

Verified by the orchestrator directly, not relayed from an agent:
- `tests` and `manifests` both **success** at `4b721b7`
- Fresh detached checkout: **24 suites, 1359 passed, 0 failed**
- The 16 suites present at `origin/main`: **531 passed, 0 failed**, unchanged

### Residuals recorded here because `.pipeline/` is gitignored and the impl-report will not survive

**Two test limits Dev named and did not reach:**
1. The QA-18 shell pin is verified on bash 3.2 (local) and bash 5 (CI), and **inferred for the bash 4.0-4.3 band**. If those versions behave like 3.2 rather than 5, that arm reddens for someone on that shell. Chosen deliberately as a loud red over a silent pass.
2. QA-20's uniqueness counter for `sha_of`/`pos_of` is controlled by observing it report `>=2` and `0` on real substrings, **not** by planting a duplicate commit subject, which would need a throwaway commit.

**Two changes Dev made outside the seven-item nit scope, flagged by Dev for dispute and accepted:**
- Round 3 was closed at `9d076c0` in `CLOSED_ROUND_TIPS`, since the nit round repairs lines round 3 introduced. Note SEC-17-N1 records that this list is hand-maintained with no earning check.
- `AC21` and `AC38` were corrected from `met: false` / `PARTIAL` to PASS. Both were stale: the same report already carried `resolved: true` for the QA repair `a7b11bb` that satisfied them. Dev re-verified both live (108/0 and 78/0) rather than trusting the report. History preserved under a `previously` key.

### Open follow-ups
#19 shared-worktree hazard, #21 unenforced config-path tier trigger, #22 corpus enumeration, #23 telemetry event-convention (buckets shifted one slot), #25 fresh-checkout CI flake.

**QA's standing condition on #25, carried forward:** if the fix gates that case behind an env var or lets it skip, that is QA-9 reopened by another door and QA will treat it as a blocker. A self-skipping case and a passing case print the same thing.

Issue #47, opened 2026-08-20
32 assertions vanish silently on ubuntu-latest because zsh is absent


nsmedia-io:
Found by Dev during #46 (2026-08-20). Same family as #46 — **a test population that shrinks silently on one platform** — but a different mechanism, so filed separately.

macOS local run: 32 suites, **1967** assertions. `ubuntu-latest` CI at the same commit: 32 suites, **1935**. Both report `failed=0` and both are green.

The 32-assertion gap is entirely in two suites neither #46 nor #39 touched:

```
test-mis-tier-tripwire.sh                78 (macOS)  ->  70 (CI)
test-panel-composition-fail-direction.sh 165        -> 141
```

Every delta line is tagged `[zsh]`, and `ubuntu-latest` has no zsh.

## Why this matters

Those `[zsh]` columns exist **because of the veto in #17**: zsh does not word-split unquoted parameter expansions, so a multi-file diff reached the surface probes as one glued argv element and the mis-tier tripwire silently failed to halt. The columns are the regression test for that veto.

**They do not run where CI runs.** So the guarantee CI reports green on is the bash half only, and a zsh-specific regression of the exact class that produced a VETO would reach `main` with a green build. The suite is honest about skipping — the lines are tagged — but the *totals* are not, and nothing compares them across platforms.

## Options

1. Install zsh on the runner (`apt-get install -y zsh` in the workflow), so the columns actually run where the gate is enforced.
2. Assert the expected per-platform assertion count so a silent shrink fails loudly, rather than being visible only to someone diffing two totals by hand.
3. Both. (1) closes the coverage gap; (2) closes the class, and would have caught this without anyone noticing the numbers.

Note the new harness count guard added in #39 catches *printed-vs-counted* mismatches within one run. It cannot catch this, because on each platform the printed and counted numbers agree — the population itself differs.

nsmedia-io:
## Measured evidence for option 2, and an argument for a third form of it

From the Phase 4 panel on #34/#42 (QA's G4). Adding it here rather than opening a fourth issue, since it is this issue's mechanism.

### The failure mode reproduces on a single platform, with no CI involved

QA deleted **one assertion** from a suite and re-ran locally:

```
66 passed / 0 failed, rc=0
```

[...]

Issue #33, opened 2026-08-19
Assertion-count floors decay silently; pin the assertion label set instead


nsmedia-io:
Raised by QA during #30 Phase 4 prep (2026-08-19), after the same defect recurred twice in two releases.

Several suites guard against silent test deletion with an assertion-count floor (`test-claims-consumers.sh`, and the pattern is copied elsewhere). The floor is a `>=` literal, and **it decays every time the suite it guards legitimately grows** — quietly, because nothing forces the literal up.

It has now happened twice:
- floor 95 against a true 96
- floor 96 against a true 99 (#30 raised the telemetry suite to 99 without deleting anything, so three assertions could have been deleted silently — the exact defect the floor exists to prevent)

**The interim patch, landed in #30:** a two-sided pin. The `>=` floor still catches deletion, and a new `<=` DRIFT ALARM fires when the suite grows past the literal, converting a silent widening into a prompt at the moment someone should be reading what was added. QA's position, quoted: the alarm is *"a patch on the symptom, not the cause"*, and it should not be removed until the real fix exists.

**The real fix:** stop counting assertions and assert the **set**. Pin a sorted digest of the suite's `ok` labels, so a deletion is identified **by name** and an addition is a diff a reviewer reads, rather than an integer nobody can interpret. A count tells you something changed; a label set tells you what.

Scope note: this touches how at least three suites report, so it is its own change rather than a fold-in. It also interacts with the fresh-checkout meta-test, which parses those same result lines.

Issue #30, opened 2026-08-19
Make three controls keep the claims they make (#16, #22, #23)


nsmedia-io:
Closes #16. Closes #22. Closes #23.

**#21 is deliberately NOT in this spec.** Ruling and reasoning are recorded as a comment on #21; it needs its own architectural run because it is the only one of the four that ADDS a halting control.

## Problem

Three controls in this repo assert a guarantee the code does not deliver. Same class the 0.21.0 release closed for the data-layer surface.

## Verified at `8c4370f` by execution, not by reading

**#16 — the migration gate cannot see an executable down section, or an empty one.**

```
executable down (THE GAP)      hasUp= true hasDown= true
commented down (correct)       hasUp= true hasDown= true
empty down                     hasUp= true hasDown= true
no down                        hasUp= true hasDown= false
```

`hasDownSection` is `downMarkerIndex(sql) !== -1` — it detects the *string* `-- DOWN`, and nothing else. `dba.md:48`, `dba.md:135` and `design.schema.json:56` all state that a down region must be commented-out documentation, never executable SQL. The gate is silent on a rule three authoritative places already state. `test-gate-pre-phase4.sh:402-417` currently asserts the bug's behaviour (`RC=0`) under a FUTURE AUTHOR note saying to INVERT it when the gap closes.

Beyond the issue: an **empty** down region also passes. `-- DOWN` followed by nothing is a down section in name only.

**#23 — telemetry credits every interval to the wrong phase.** Re-derived from the real record at `.pipeline/17/status.json`, both conventions computed side by side:

| phase | current parser (entry) | corrected (exit) |
|---|---|---|
| `0.5-map` | 563676 | 6545 |
| `1-ba` | 819137 | 557131 |
| `2-review` | 1084949 | 819137 |
| `1-ba-rework` | 605812 | 1084949 |
| `2-review-r2` | 1771108 | 605812 |
| `1-ba-rework-r2` | **0** | 1771108 |
| `2.5-design` | 1532139 | **0** |
| `3a-qa-tests` | 4088488 | 1532139 |
| `3b-dev` | *(no key)* | 4088488 |

**All 10 real events carry a `verdict`** (verified: `events with a verdict field: 10 / 10`). A verdict is not knowable before a phase runs, so these are exit markers. Two independent confirmations that exit is the true convention:

[...]

PR #51, opened 2026-08-20
Bound the archived verdict fields, and de-rot the phase list in status.schema.json

### LINKED ISSUE #34: events[].verdict has no length cap, in a file that is committed and archived verbatim
Carried forward from the #30 Phase 2 review and re-raised by SecOps in the #30 Phase 4 panel as a low-severity, non-blocking finding. Deferred out of #30's fix round, written here per the deferral rule (an item routed to a follow-up must be written in that issue, with its evidence, before the change that deferred it merges).

## The finding

`plugins/pipeline/schemas/status.schema.json`, `events.items.properties.verdict`, is still:

```json
{"type": "string"}
```

No enum, no `maxLength`. `status.json` is committed to git and archived verbatim by the Librarian, so whatever an agent writes into that field lands in a public tree at full length.

#30 raises the stakes on this specific field rather than leaving it neutral: it promotes `verdict` to the documented tell that fixes the meaning of a stored measurement. `plugins/pipeline/scripts/pipeline-telemetry.mjs`'s docstring now reads "The tell is `verdict`", and `plugins/pipeline/commands/pipeline.md:67` states that `events[]` entries are EXIT markers carrying the verdict of the phase they close. A field that decides how a stored number is read is a field worth bounding.

## Precedent in the same file

`flags[].summary` already carries `maxLength: 140`. The mechanism exists; it was simply never applied here.

## Recommended fix: a cap, not an enum

SecOps' reasoning, recorded so it is not re-litigated:

- An **enum** refuses a novel verdict. Verdict vocabularies grow (`APPROVE_WITH_NOTES` was itself an addition, and `SECOPS_VETO` is role-specific), so an enum turns a legitimate new verdict into a schema failure at exactly the moment someone is trying to record it.
- A **`maxLength`** refuses only the free-text prose the numbers-only convention already forbids. It cannot refuse a correct verdict, because no correct verdict is long.

So: `{"type": "string", "maxLength": <n>}`, with `n` chosen against the longest verdict currently in use plus headroom.

## Scope note, from the same review

`events[].phase` remains a bare string too. That was the materially risky one because it was KEY-FORMING, and #30 closed that: `pipeline-telemetry.mjs:181` keys on `phaseKey(...)` rather than the raw label, verified by SecOps against 40 hostile labels (path-shaped, `__proto__`/`constructor`, 5000-char, secret-shaped, control chars) with 0 keys outside `KNOWN_PHASES` and 0 prototype pollution. `phase` is therefore NOT part of this issue; it is named here only so the next reader does not re-derive it as an open hole.

## Where it came from

- #30 `.pipeline/30/peer-review.secops.json`, `concerns[]`, severity `low`, location `plugins/pipeline/schemas/status.schema.json:events.items.properties.verdict`.
- Raised at #30 Phase 2 and not taken; re-raised at Phase 4 with the explicit condition that deferring requires this issue to exist before #30 merges.

### LINKED ISSUE #42: status.schema.json:13's description example list has rotted (blesses two phases pipeline.md never writes)
Follow-up filed at Phase 4 peer review of #39, closing a deferral that spec.json (rev 2) claimed was "filed as a follow-up" but for which no issue actually existed until now.

## The defect

`plugins/pipeline/schemas/status.schema.json:13`'s `current_phase` description carries a hand-maintained example list of phase names. That list has already rotted: it blesses `3-scope-drift-adjudication` and `3-impl-verification-unverified`, and VERIFIED neither appears in `commands/pipeline.md` (which writes `3-impl-live-verify-unverified` instead).

```
grep -n 'scope-drift-adjudication\|verification-unverified' plugins/pipeline/commands/pipeline.md
# no match
```

This is the fourth member of the "hand-maintained phase vocabulary rots" class named in #39's spec (`sibling_causes_considered`), alongside `status.schema.json`'s regex pattern, `dispatch-model.mjs`'s `KNOWN_PHASES`, and `voice-lint.mjs`'s `VOICE_MOMENTS`/`NON_VOICE_PHASES`. It is the one member that rotted **in the file the drift test does not read**, which is exactly why nobody noticed until #39's rev-2 review pass found it by direct inspection.

## Why it was cut from #39 rather than fixed there

Per #39's `spec.json` `out_of_scope`: it is a docs edit in a schema file with no behavioural coupling to the phase-entry guard. `gate-phase-entry.mjs`'s R13/AC14 already define runtime behavior for a `current_phase` the guard doesn't recognize (`not-applicable`, never `refused`), and AC19 already asserts a second ground truth derived from the committed `.pipeline/*/status.json` corpus rather than from any prose table — so a phase like this reaching a real record is independently caught. Fixing #39's own diff to also correct this description would have widened a one-mechanism change.

## What "fixed" looks like

Update `status.schema.json:13`'s description to only name phase literals that `commands/pipeline.md` actually writes (the drift test in #39's `tests/test-gate-phase-entry-drift.sh` derives that list mechanically via `grep -o 'current_phase: *"[^"]*"'` and could plausibly be reused or referenced here, though this schema description is prose, not machine-enforced, so there's no requirement that the fix itself be machine-checked -- just accurate at time of edit).

## Evidence trail

- `#39` spec.json `rulings.R13` and `out_of_scope` (BA, rev 2)
- `#39` review.dba.json blocker 2 (DBA, who found this)
- `#39` `.pipeline/39/impl-report.json` AC14/AC19 (Dev, confirming the guard's handling of an unrecognised phase is independent of this fix)

Filed by BA at Phase 4 peer review of #39, because the deferral had been described as already done ("filed as a follow-up") in spec.json's `ba_notes` without a real issue existing -- a repeat of a known failure mode in this project (a spec claiming "Filed." for an issue that did not exist). #39's PR (#41) is still open at filing time.

nsmedia-io:
Two edits to `plugins/pipeline/schemas/status.schema.json`, plus the test that makes the first one load-bearing.

## What was wrong

`events[].verdict` and `flags[].verdict` were bare `{"type": "string"}` in a file that is committed to git AND archived verbatim into the knowledge store, so any prose an agent wrote into them reached a public tree at full length. Separately, `current_phase`'s hand-maintained example list had rotted: it named two phases `commands/pipeline.md` never writes, and omitted six it does.

Both are the same underlying condition. `status.schema.json` has exactly one runtime reader — `voice-lint.mjs:224`, which reads `properties.current_phase.pattern` and nothing else — so a constraint written into this file can be wrong or absent indefinitely with nothing going red.

## Correcting #34's premise

#34 says of `flags[].summary`'s `maxLength: 140` that "the mechanism exists; it was simply never applied here." It does not exist. `status.schema.json` appears in no `AGENT_RULES` entry in `validate-pipeline-artifact.mjs`, and that walker does not implement `maxLength` at all; `voice-lint.mjs:238` puts it in one sentence — "Nothing else validates this file." The 140-cap is honored only because `pipeline.md` restates it as a writer rule.

So a `maxLength` shipped alone here would be a comment, not a guard. Two consequences in this PR: each capped field's `description` now says outright that nothing validates `status.json` against this schema, so a later reader does not mistake the cap for an enforced control and skip writing the enforcement; and a corpus-conformance test makes the cap load-bearing against the population #34 actually cares about — the committed records that reach the public tree.

## The changes

**The cap: 32 on both verdict fields.** Derived, not round: 18 from the longest member of the DECLARED verdict vocabulary across all four artifact schemas (`APPROVE_WITH_NOTES`), plus headroom for any longer SCREAMING_SNAKE token, sitting below the length of the prose sentence the bound actually refuses. `flags[].verdict` is capped alongside `events[].verdict` rather than deferred: #34 names only `events[]`, but `flags[]` is the same field name in the same file under the same archive-verbatim mechanism and carries the LONGER of the two observed values (18 vs 15), so deferring would have left the bigger half open while closing the smaller. Zero of 6 committed records and zero of 26 verdict-carrying entries are retroactively invalidated. An absent `verdict` stays valid — the cap must not become a requirement, and one live record depends on that.

**The phase list.** Now exactly the literals `pipeline.md` writes, plus `halted-error`, with the `<phase>-error` family named as a SHAPE rather than enumerated. `halted-error` is the trap: `pipeline.md` writes it nowhere, so a list mechanically derived from `pipeline.md` alone deletes it — yet the schema's own `pattern` alternates it and `gate-phase-entry.mjs:241` is `if (phase === "halted-error") return true;`. Removing it would refuse correct work, and nothing would have caught it, because the guard does not read the description.

**`properties.current_phase.pattern` is untouched** and pinned byte-identical, since it is the one key the schema's only runtime reader consumes.

**Three test comments re-anchored.** `test-gate-phase-entry-drift.sh:13`, `test-gate-phase-entry.sh:511` and `:711` cited the rotted list as LIVE present-tense evidence. All three became false on landing, so they are restated as history naming #42. The negative-control INPUT at `test-gate-phase-entry.sh:514` is byte-unchanged: `pipeline.md` still writes `3-scope-drift-adjudication` nowhere, so it remains a phase the guard can genuinely meet and must not refuse.

## The new suite

`plugins/pipeline/tests/test-status-schema-contract.sh`, 67 assertions. It reuses the existing `#30` union corpus helper rather than building a second population.

- **The cap is READ FROM the schema** at every corpus assertion. Exactly one site holds the literal `32`, as the pin on the ruled value. A test carrying its own copy would stay green while the schema drifted away underneath it.
- **A vacuity control runs before any property of the corpus is asserted.** The floor is derived from two independent enumerations (git's index and the filesystem) and is never a pinned integer — `origin/main` holds 5 committed records and this branch holds 6, so a pinned `6` would measure the branch rather than the repo. A walk that returns zero FAILS; it does not pass green.
- **A permanent non-zero control** drops the cap to 17 and requires the check to go red and NAME `.pipeline/17/status.json`'s `APPROVE_WITH_NOTES`. It carries its own expiry in the assertion message, because it is anchored to a live record and the correct fate of a live record is that somebody edits it.
- **Set equality in both directions** against `pipeline.md`, with a control per direction, plus vacuity floors on both sets.

## Evidence

Twelve mutations, applied after the commits landed and restored from git, each printing the changed line as proof it applied (one silently failed to apply on the first attempt and was caught by exactly that step):

| Mutation | Result |
|---|---|
| cap 32 → 17 (schema only) | RED: `.pipeline/17/status.json flags[4].verdict="APPROVE_WITH_NOTES" len=18 cap=17` |
| delete `flags[].verdict` cap only | RED on the flags pin, events pin green |
| delete `events[].verdict` cap only | RED on the events pin, flags pin green |
| add a bogus literal to the list | RED, direction 2 |
| delete a real literal from the list | RED, direction 1 |
| remove `halted-error` | RED (this is the mutation the naive fix performs) |
| re-insert `3-scope-drift-adjudication` | RED, file-scoped |
| `pattern` `[0-5]` → `[0-6]` | RED, byte comparison |
| restore a comment to present tense | RED, names the file:line |
| delete one assertion from the suite | 67 → 66 with `failed=0` — #47's failure mode, caught only by the count |

**One mutation SURVIVES, deliberately:** raising the cap 32 → 64 leaves the corpus-conformance assertion GREEN. It should. The corpus maximum is 18, so no committed record can distinguish 32 from 64 — AC3 constrains the cap from below (via the 17 control) and not from above. A battery in which every mutation reddens cannot tell coverage from a rubber stamp.

`bash plugins/pipeline/tests/run.sh`: **2056 assertions, failed=0, exit 0.** Pre-change was **1989** at `4a6f164`; both measured on the same macOS host (Darwin 25.5.0 x86_64), per #47's real 32-assertion macOS/ubuntu gap. The +67 delta is exactly the new suite's standalone count, so no pre-existing suite lost an assertion. The numbers are recorded in `impl-report.json` with their platform and population rather than pinned into a test, per #33.

`test-voice-lint.sh` (the only runtime consumer's suite) is green at 41/0, and its non-zero control was run rather than assumed: a malformed `current_phase` still exits 2 and still names `status.schema.json`, and well-formed phases stay silent.

## Known limits, stated rather than implied

- The corpus floor does not catch a record deleted from BOTH the index and the disk; both enumerations fall together. The alternative is a pinned integer, which decays (#33) and here would measure the branch.
- The cap is still unenforced at WRITE time. Registering `status.json` in `AGENT_RULES` is a new enforcement surface with its own fail-open/fail-closed design, and implementing `maxLength` in the walker would change every schema at once — both out of scope. What this buys is that such a value cannot survive a run of the checkCommand unnoticed.
- The new suite exempts its own file from the present-tense grep, because it must carry both the needle and the pre-change wording to check them at all.

Closes #34
Closes #42

### DIFF
--- a/.pipeline/34/status.json
+++ b/.pipeline/34/status.json
@@ -2,9 +2,16 @@
   "issue_number": 34,
   "current_phase": "1-ba",
   "started_at": "2026-08-20T13:32:41Z",
-  "updated_at": "2026-08-20T13:32:41Z",
+  "updated_at": "2026-08-20T13:37:14Z",
   "branch": "claude/mechanical-group-impl-bfb9ce",
   "ask_text": "Bound events[].verdict with a maxLength in status.schema.json (#34) and correct the rotted current_phase description example list at line 13 (#42). Both are single-file edits to plugins/pipeline/schemas/status.schema.json.",
-  "events": [],
+  "events": [
+    {
+      "phase": "0.5-map",
+      "verdict": "SKIPPED",
+      "at": "2026-08-20T13:37:14Z",
+      "note": "No separate 0.5-map dispatch: pipeline.md folds the map into the Phase 1 BA dispatch below the architectural tier, and BA (in flight) writes map.json alongside spec.json. Recorded as a skip rather than waited on because this record carries no risk_tier yet -- the tier is BA output, while pipeline.md mandates checkpointing 1-ba BEFORE dispatching BA -- so gate-phase-entry.mjs:133 normalizes the absent tier to architectural and switches on the architectural-only 1-ba row. That is issue #45 reproduced live. Revisit if BA promotes the tier, which needs the deep three-layer map as its own dispatch."
+    }
+  ],
   "flags": []
 }
--- a/.pipeline/34/status.json
+++ b/.pipeline/34/status.json
@@ -1,8 +1,8 @@
 {
   "issue_number": 34,
-  "current_phase": "2-constraints",
+  "current_phase": "4-review-complete",
   "started_at": "2026-08-20T13:32:41Z",
-  "updated_at": "2026-08-20T13:47:36Z",
+  "updated_at": "2026-08-20T16:34:45Z",
   "branch": "claude/mechanical-group-impl-bfb9ce",
   "ask_text": "Bound events[].verdict with a maxLength in status.schema.json (#34) and correct the rotted current_phase description example list at line 13 (#42). Both are single-file edits to plugins/pipeline/schemas/status.schema.json.",
   "events": [
@@ -16,6 +16,22 @@
       "phase": "1-ba",
       "verdict": "complete",
       "at": "2026-08-20T13:47:36Z"
+    },
+    {
+      "phase": "2-constraints",
+      "verdict": "complete",
+      "at": "2026-08-20T13:48:59Z",
+      "note": "dba 12 / devops 11 / secops 13 lines, each verified non-empty separately; extracted from the repo agents dir, not the stale plugin cache."
+    },
+    {
+      "phase": "3-impl",
+      "verdict": "complete",
+      "at": "2026-08-20T15:02:13Z"
+    },
+    {
+      "phase": "4-review",
+      "verdict": "APPROVE_WITH_NOTES",
+      "at": "2026-08-20T15:50:58Z"
     }
   ],
   "flags": [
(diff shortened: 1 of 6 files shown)


nsmedia-io:
## Phase 4 Peer Review (standard tier panel)

|---|---|---|
| BA | APPROVE_WITH_NOTES | none — 2 notes (BA-1, BA-2) |
| Dev | APPROVE_WITH_NOTES | none — 5 notes (D1–D5) |
| QA *(binding test verdict)* | APPROVE_WITH_NOTES | none — 1 demonstrated hole (G1) + 3 gaps |
| SecOps *(veto held, not exercised)* | APPROVE_WITH_NOTES | none — 2 medium, 1 info |

Not on panel (standard tier): DBA, DevOps — both withheld on a probe returning the reserved no-match code, not on an unevaluable probe. Design not seated (no frontend surface). Art Director not seated (no visual contract).

**Final verdict: APPROVE_WITH_NOTES** · `review_rounds: 1`

### Verified by the orchestrator rather than relayed

- Suite re-run independently: **2056 assertions / 33 suites / failed=0 / exit 0**, matching `impl-report.json`.
- Cap mutation planted by hand (schema `32 → 17`, **zero test files touched**): suite went **67/0 → 63/4**, AC3 naming `.pipeline/17/status.json flags[4].verdict="APPROVE_WITH_NOTES" len=18 cap=17`. That is AC3 and AC4 proven in one stroke.
- Both pre-Phase-4 gates: artifact gate `OK`, frontend gate `SKIP`.
- Issue #52 confirmed to exist (SecOps' deferral was written, not claimed).

### Two findings reached independently by two lenses

**The `0-setup` divergence** (Dev D1, BA BA-2). This PR adds a *third* hand-written derivation of the phase vocabulary using a loose grep (27 tokens / 26 literals) where the two existing ones are strict (26 / 25). The sole difference is `0-setup`. Consequence: AC5's set equality now requires `0-setup` in the schema description, making the schema the only authority in the repo naming it — it is absent from `voice-lint.mjs`'s `VOICE_MOMENTS`/`NON_VOICE_PHASES` and from `gate-phase-entry.mjs`'s tables. `test-voice-lint.sh:234` asserts *"every phase pipeline.md writes is accounted for in voice-lint.mjs"* **and passes**, because its narrow derivation never sees the missing member.

Sharpened by the comment this same PR writes into `test-gate-phase-entry-drift.sh`: *"a hand-maintained second copy of a vocabulary rots silently unless something compares it to the first."* This change adds the copy without the comparison.

**AC2's floor is effectively `>= 1`, not `>= 6`** (QA, BA). A corpus collapsing 6 → 1 passes green. The two lenses agree on the fact and differ on remedy: BA proposes pinning 4 (the AC19 precedent the spec itself cites); QA judges it acceptable because a different criterion re-anchors it.

### QA's demonstrated hole (G1)

`test-status-schema-contract.sh:447-458`. AC10 walks all suites but guards itself with a **pinned** floor of `>= 8` against a real population of 33.

- Planted the forbidden wording in `test-voice-lint.sh` → **66/1**, AC10 red. Needle alive.
- Same violation, glob narrowed to 9 files (floor satisfied) → **67/0 green. Violation unseen.**

This is the new `evidence.md` rule recursing into the diff that cites it — *ask what THAT guard fails to bind.* The diff derives AC2's floor correctly and then pins this one. Reachable only by a future edit, so a ratchet rather than a live gap.

### Where the change's own rationale does not hold

The shipped description justifies the cap as *"status.json is committed AND archived verbatim, so an unbounded string here reaches a public tree at full length."* Measured on the same corpus, the longest `events[].note` is **616 characters** (`.pipeline/39/status.json`) — 19× the verdict cap, same file, no cap at all. SecOps ruled a length cap is the **wrong instrument** there (the 600-char note in this very PR is correct audit work) and that the real dimension is content, not length — filed as #52.

Separately, SecOps' **S1**: the description claims the cap is *"honored by the writer"*, but `commands/pipeline.md` restates the 140-char `summary` rule and restates the verdict cap nowhere. The sibling description eight lines below correctly names its own mechanism, so the asymmetry reads as symmetry.

### SecOps' central ruling

**Detects, does not prevent — and that is sufficient for the issue as filed.** Nothing validates `status.json` at write time (`maxLength` appears once in `validate-pipeline-artifact.mjs`, inside the comment declaring it advisory). But the corpus check runs in CI `on: pull_request`, before the merge into the public tree and before Phase 5 archival, so event and environment match. Residual: when CI reddens, the string is already in the branch's history, and nothing tells the author to amend rather than fix forward.

Issue #46, opened 2026-08-20
AC23(a)/AC28(a) fixtures don't construct the missing-node condition on Linux, so the fail-open posture is ungraded where CI runs


nsmedia-io:
Found by Dev during the #39 Phase 4 fix round (2026-08-20). **This is why the branch is CI-red**, and it predates that round — the same two assertions failed at `2c6c70c`.

## The defect

`hooks/stop.sh:21` unconditionally runs:

```bash
export PATH="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:$PATH"
```

so the hook **rebuilds a PATH the test just stripped**. On `ubuntu-latest` the runner's bundled Node is at `/usr/local/bin/node`, so the guard is armed after all: it refuses with rc 2 (test expects 0), and *because* it is armed, `session-start.sh` correctly emits no disarm notice (test expects 1). One root cause, both failures.

It passes on macOS for a reason that is a fact about one machine rather than a property of the fixture: `command -v node` resolves under `~/.nvm/...`, which is absent from all four prepended directories, and the fixture also redirects `HOME`, killing stop.sh's nvm branch.

## The assertion defect underneath it

The precondition at `test-gate-phase-entry-hook.sh:190-192` measures `command -v node` under `env -i HOME=… PATH=…` — **the PATH the test builds, before `stop.sh` mutates it.** What it inspects is not what acts, which is why the cell fails with a green precondition directly above it.

## Why it matters beyond a red build

Direction is **claims-more, on the platform that matters**. On Linux that cell has never exercised a missing-node path, so the **fail-open-on-tooling posture SecOps explicitly signed off on in #39 is ungraded on the platform CI actually runs.** The macOS pass was never evidence for the Linux behaviour.

## Options

1. **Fix the fixture** — construct the condition through the hook's *effective* PATH (shadow node in the prepended positions), or make the precondition assert through that PATH so it fails loudly wherever the strip does not hold. Test-only; does not touch the hook's posture.
2. **Drop the hardcoded prepend in `stop.sh:21`** — but it exists to defeat a stale app-inherited nvm Node, so removal has its own cost and changes a posture SecOps signed off on. That trade belongs to BA and SecOps, not to a fix round.
3. Defer.

Dev recommended option 1 and flagged it as a scope question rather than fixing it unasked. Note that "fail loudly" alone leaves CI honestly red; **shadowing actually constructs the condition**, which is the version that makes the assertion mean what it says on both platforms.

[...]

Issue #44, opened 2026-08-20
spec.json contradicts itself on AC12: expected_survivor forbids the exact assertion AC12's text mandates


nsmedia-io:
Found by QA in #39 Phase 4 (2026-08-19). This is a BA-domain defect: two sections of one spec disagree, and the suite followed the wrong one.

`spec.json → falsifiability_pass.expected_survivor` says in capitals that AC12/13/14 **"must assert on the REASON STRING, never on the exit code"**, and that catching that survivor via an exit-code test means *"that test is asserting something other than what it claims."*

`spec.json → acceptance_criteria` AC12 then says only: *"...yield `not-applicable`, and stop.sh exits 0."*

The suite followed the criterion text and wrote `assert_eq "  and stop.sh's side of it exits 0" "$GATE_RC" "0"` — literally the forbidden assertion.

**Measured consequence.** Mutations **M12a, M12c, M12d all SURVIVED**: the entire `isTerminal` phase-name check can be deleted with zero failures, because every terminal phase is also absent from `GUARDED`, so the vocabulary fail-open returns `not-applicable` too. **M12e SURVIVED** as well — changing the reason from `"is finished."` to `"WRONGREASON."` breaks nothing.

Dev's own notes concede the related point: AC13 *"cannot tell unguarded from unknown."*

Fix is to reconcile the two sections and then assert on the reason string, so `terminal` and `unguarded` are distinguishable — they are different facts about a run and currently produce an identical observable.

nsmedia-io:
**Handoff — this is not in `stop.sh`, it is in `tests/test-gate-phase-entry.sh`.**

The lane board routed this to Lane 3 as "same file as #46 (`stop.sh`)". That premise is wrong. The forbidden assertion the issue quotes is at [`tests/test-gate-phase-entry.sh:486`](https://github.com/nsmedia-io/agent-pipeline/blob/main/plugins/pipeline/tests/test-gate-phase-entry.sh#L486) — the gate suite, which Lane 1 owns — so Lane 3 is handing it over rather than reaching across.

**Still live at `4fffbdd`, measured, not restated.** Both mutations were run against an unmodified checkout with `SCRIPTS_DIR` pointed at a mutated copy of `scripts/`:

| mutation | result |
|---|---|
| `is finished.` → `WRONGREASON.` (M12e) | `passed=329 failed=0` — **SURVIVES** |
[...]

Issue #25, opened 2026-08-19
AC41(c) fresh-checkout CI case is flaky, cause unknown


nsmedia-io:
Found and left OPEN during the #17 veto-fix round (2026-08-19). Recorded here because the branch artifact that carries it (`.pipeline/17/impl-report.json`, `ci.known_flake`, task `T-R3-6`) is gitignored and would not survive the merge.

`test-issue17-integration.sh`'s AC41(c) case runs `run.sh` against a fresh checkout, so that "the suite passes" stops meaning "in the worktree where it was written".

**It is flaky.** Run `32256552933` (head `788bc22`) failed it. A rerun of the **identical tree** passed. One suite inside the clone went red and the case could only report "23 of 24 reported failed=0", with no way to tell which.

Mitigation shipped, cause not found: the case now names the failing inner suite, echoes its FAIL lines, and carries a two-sided control on the extractor, so the next occurrence describes itself.

`tests` is green at `9d076c0` and in a local fresh checkout (24 suites, 1314 passed, 0 failed), verified independently. **Do not read that green as proof this is settled.** A gate that reddens without saying why is the gate that eventually gets switched off, which is the argument the workflow's own header makes.

Next step when it recurs: capture the new diagnostic output and identify the inner suite. Likely candidates are ordering or a shared temp path between the outer run and the inner clone.

nsmedia-io:
**#25 and #27 are the same flake.** Observed together on the 0.21.0 release PR (#28), run `32277458463`, on a version-bump-only tree:

```
FAIL  run.sh exits 0 in the fresh checkout
FAIL  and says so
FAIL  and every one of them reported zero failures
FAIL  node validate-pipeline-artifact.mjs --self-test exits 0
FAIL  the self-test reports zero failures
```

The fresh-checkout case reddens **because the self-test flakes inside the clone**. The diagnostics added in the nit round worked as intended: this case now names its inner cause instead of only reporting "23 of 24 reported failed=0".

**Two things ruled out, so nobody repeats them:**
- `./harness.sh: line 41: printf: write error: Broken pipe` in the same log is a **red herring**. Line 41 is the FAIL-reporting path (`printf %s "$haystack" | head -3`), so the broken pipe is a consequence of an assertion already having failed, not the trigger.
- It does **not** reproduce standalone: `validate-pipeline-artifact.mjs --self-test` run 8 consecutive times locally gave 8× `61 passed, 0 failed`, rc 0.

[...]

PR #65, opened 2026-08-21
Test-population integrity and CI parity (#47, #37, #33)

### LINKED ISSUE #33: Assertion-count floors decay silently; pin the assertion label set instead

### LINKED ISSUE #37: Ratchet: a test population must never be derived from a range against a moving ref
Third occurrence of one shape, twice breaking `main` at merge.

**Occurrence 1** (#17 merge): `test-issue17-integration.sh` computed its window as `git merge-base origin/main HEAD` .. `HEAD`. On `main` that merge-base *is* HEAD, so the range was empty, every subject lookup returned `""`, and `set -u` killed the suite before it printed a count. Fixed in #26 by delimiting the series on commit **subject** and resolving against full history.

**Occurrence 2** (#32 merge): `test-claims-consumers.sh:187` built AC30's population from `git log --format=%s origin/main..HEAD`. Empty on `main`, so the control asserting there were subjects failed and the fresh-checkout meta-test cascaded (5 failures total). Fixed in #36.

**The lesson did not generalise, because nothing enforces it.** Occurrence 2 was written three review rounds *after* occurrence 1 was fixed, in the same repo, by agents who had the first fix in front of them. A rule that lives only in a commit message is a rule the next author does not have.

## What to build

An assertion that fails when a test file derives a **population over the real repository** from a range against a ref that moves — `origin/main..HEAD`, `origin/main...HEAD`, `merge-base origin/main HEAD`, or equivalent.

Two things make this non-trivial, and both must be handled or the ratchet is worse than nothing:

1. **Throwaway test repos legitimately use these ranges.** `test-mis-tier-tripwire.sh`, `test-panel-composition-fail-direction.sh` and others construct a local `origin/main` ref inside a `mktemp` repo precisely so the shipped block's own `git diff origin/main...HEAD` resolves. Those are correct and must not be refused. The discriminator is whether the command runs against `$REPO_ROOT` or against a constructed fixture repo.
2. **Comments and heredoc fixtures quote these strings.** The `#30` work already hit the self-counting trap twice (a detector matching its own fixture, and a docstring assertion counting its own prose). Whatever this ratchet greps for, it must not count itself.

Reference implementations in-repo: `test-issue17-integration.sh`'s `same_local_offenders` (a scanner that deliberately skips comments and quoted fixtures, with 1-and-0 controls on both sides) and the AC22 docstring-window derivation (bounds read from the file rather than pinned to a line count).

## Gate-bites requirement

Plant the exact line from occurrence 2 (`git log --format=%s origin/main..HEAD` against `$REPO_ROOT`) and record the ratchet catching it. Then plant a legitimate fixture-repo use and record it passing. A ratchet that refuses both is a ratchet that gets deleted the first week.

### LINKED ISSUE #47: 32 assertions vanish silently on ubuntu-latest because zsh is absent

nsmedia-io:
Lane 3. Three fixes, one theme: **CI was reporting green on guarantees it did not check, and the guards meant to notice were counting instead of naming.**

Closes #47, closes #37, closes #33.
(#46 was already fixed on `main` and has been closed separately with its CI evidence. #44 turned out to live in Lane 1's file and has been handed over. #25 stays open by design.)

---

### #47 — 32 assertions did not exist where the gate is enforced

The `[zsh]` columns in `test-mis-tier-tripwire.sh` and `test-panel-composition-fail-direction.sh` are the regression test for the #17 **VETO**: zsh does not word-split an unquoted parameter expansion, so a multi-file diff reached the surface probes as one glued argument and the mis-tier tripwire silently failed to halt. `ubuntu-latest` has no zsh. Those columns ran on a laptop and nowhere else — 70 and 141 assertions in CI against 78 and 165 locally, both printing `failed=0`, both green.

Both options in the issue:

1. The runner installs zsh.
2. The columns are **declared**, not discovered. Each suite opened its column with a bare `command -v zsh` and closed the else branch with `assert_eq "..." "not-run" "not-run"` — an assertion that cannot fail, printed in the same `ok` column as one that can. `harness.sh` grows `optional_tool`, which records the tool and its state either way and, under `PIPELINE_TESTS_REQUIRE_CAPABILITIES=1` (set by the workflow and nowhere else), converts an absent tool into a counted failure **naming it**.

The issue's second option was an asserted per-platform assertion count. A count is the instrument #33 is open against — it decays every time the suite it guards legitimately grows. A capability name does not decay. With zsh installed the population no longer depends on the platform at all, which is what lets #33's pin be stated once rather than once per platform.

**Gate bites**, on a PATH stripped of zsh:

| | result |
|---|---|
| strict + absent | rc 1, `` FAIL CAPABILITY `zsh` is present ``, `passed=69 failed=1` |
| ordinary + absent | rc 0, the absence named, `passed=70` — exactly CI's number |
| strict + present | rc 0, `passed=79` |

### #37 — the ratchet

Third occurrence of one shape, twice breaking `main` at merge. `tests/test-moving-ref-population.sh` walks every file `run.sh` can reach and refuses a range against a remote-tracking ref pointed at **the checkout** — named by `$REPO_ROOT` and friends, or pointed at nothing, which leaves the command in the suite's own cwd. It names the offender by file and line.

Both hard halves are handled and asserted:

- **Fixture repos legitimately use these ranges.** The gate-bites pair the issue asks for is recorded: occurrence 2's exact line is caught, and the identical range pointed at a fixture repo passes.
- **Comments and heredocs quote these strings.** Full-line comments are prose; the uncommented twin of the same text is caught. The detector's own remote-ref half is assembled at run time, so this file sits inside the population it walks and passes by construction rather than by exemption.

### #33 — pin the set, not the count

AC19's `>=` floor decayed twice (95 against a true 96, then 96 against a true 99, at which point three assertions could have been deleted silently *through* the guard). #30's `<=` DRIFT ALARM was the interim patch, kept until the real fix existed. It exists: AC19 now pins the **set of assertion labels** each guarded suite emits, as tracked files under `tests/fixtures/labels/`. A deletion is reported by the name of the assertion that vanished. Four literals are gone.

**Gate bites:** deleting `assert_eq "an AC-label match covers the criterion" ...` reddens AC19 with that label quoted back — where the floor would have gone 95 → 94 and said nothing.

Both label sets were verified byte-identical between macOS and `ubuntu-latest` before the pin was taken, and every sort and `comm` is `LC_ALL=C`.

### Folded in (rule 0.1 — same file, no tier change, recorded rather than filed)

- `test-claims-consumers.sh` AC30's #36 patch left a control that **could not fail**: its `elif` re-ran the identical command whose emptiness had already put it there, so both branches answered `ok`. AC30 now asks its question of full history only, which is strictly stronger, with two computed non-zero controls.
- `test-issue17-integration.sh`'s `npm install` / `npm ci` checks matched the **whole workflow file**, so a comment explaining that CI installs no npm dependency turned them red while no npm ran. They now read the `run:` steps, with a control that the extraction found any and a planted npm step proving the reader still bites. The zsh step and the strict-capability env are pinned there too, so dropping either reddens a named assertion.

---

**Verification.** Full suite locally: 34 suites, `failed=0`, including AC41(c)'s fresh-checkout case. CI is the remaining unknown this PR exists to settle — specifically that `apt-get install -y zsh` succeeds and that the two suites then report 79 and 165 rather than 70 and 141.

**Net issue count for this lane: −4 closed, 0 opened.**

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,9 +27,34 @@ jobs:
           # This job is the only place the depth matters, and a repo of markdown and JSON is
           # small enough that a full fetch costs nothing worth measuring.
           fetch-depth: 0
-      # No package.json, no dependencies, no install step, matching manifests.yml's own
+      # THE ONE INSTALL STEP, and it is here because leaving it out was measured to cost 32
+      # assertions. test-mis-tier-tripwire.sh and test-panel-composition-fail-direction.sh each
+      # carry a `[zsh]` column, and that column is the regression test for the #17 VETO: zsh does
+      # not word-split an unquoted parameter expansion, so a multi-file diff reached the surface
+      # probes as one glued argument and the mis-tier tripwire silently failed to halt. zsh is
+      # the shell the orchestrator's own tool runs. ubuntu-latest has no zsh, so those columns
+      # ran on the author's laptop and NOWHERE ELSE -- 70 and 141 assertions here against 78 and
+      # 165 locally, both reporting failed=0, both green (#47). The gate was enforced on the
+      # bash half only.
+      #
+      # This does not reopen what the note below refuses. The constraint there is about a gate
+      # that breaks on SOMEONE ELSE'S package -- a lockfile, a registry, a transitive dep that
+      # moves. zsh is a distro package this workflow names by hand and pins to nothing; if apt
+      # cannot produce it, the honest outcome is a red build, because the alternative is the
+      # silently-narrower suite that made this step necessary.
+      - name: install zsh (the [zsh] columns are the #17 veto's regression test)
+        run: sudo apt-get update && sudo apt-get install -y zsh
+      # No package.json, no dependencies, no npm install step, matching manifests.yml's own
       # deliberate constraint: a gate that can break on someone else's package is a gate that
       # gets disabled the first time it does. run.sh is bash plus the runner's bundled Node.
       # A non-zero exit from run.sh fails the build; that is the whole point of the job.
+      #
+      # PIPELINE_TESTS_REQUIRE_CAPABILITIES turns an absent optional tool from a column that
+      # quietly does not exist into a counted failure naming the tool (see optional_tool in
+      # tests/harness.sh). It is set HERE and only here: a developer's machine may legitimately
+      # lack zsh, the platform the gate is enforced on may not. Without it, uninstalling zsh
+      # above would put this job straight back to a green build over a narrower suite.
       - name: plugin test suite
+        env:
+          PIPELINE_TESTS_REQUIRE_CAPABILITIES: "1"
         run: bash plugins/pipeline/tests/run.sh
--- a/plugins/pipeline/tests/fixtures/labels/test-gate-pre-phase4.labels
+++ b/plugins/pipeline/tests/fixtures/labels/test-gate-pre-phase4.labels
@@ -1,0 +1,95 @@
+BOUNDARY: a MySQL conditional-execution opener counts as an up section
+CONTROL: a real paraphrase of the same long criterion still covers it
+CONTROL: the same criterion passes once one check names AC23
+CONTROL: the same statements unwrapped ARE an up section
+GUARANTEE: an executable (uncommented) down region now halts
+NOT as commented-out: a `*/` past the marker must not vouch for the region
+a block that opens before the marker and closes after it halts
+a block-comment PREAMBLE above live statements is still an up section
+a block-commented up region is NOT an up section
+a custom glob matches ONLY what it names
+a custom glob matches what it names
+a file using only the configured custom marker passes
+a labelled criterion no check names is uncovered, despite heavy token overlap
+a labelled criterion still matches on wording when NO check carries a label
+a migration recorded in top-level files_removed is exempt
+a migration with no down section halts
+a migration with no up section halts
+a non-array migrationGlobs falls back to the defaults
+a non-string glob element falls back to the defaults
+a non-string marker falls back to the default
+a passing gate says so
+a per-commit exemption does not leak to a later commit
+a per-commit files_removed exempts that commit
+a schema violation halts
+a short criterion still matches on three tokens, exactly as before
+a token-overlap match covers the criterion
+a whitespace-only marker falls back to the default
+an AC-label match covers the criterion
+an absent config applies the default globs
+an absent impl-report.json halts
+an absent spec.json halts
+an all-comment head is not an up section
+an empty marker still halts a down-less migration
+an empty requirement_checks halts (minItems)
+an uncovered criterion halts the panel
+an unparseable config does not crash the gate: defaults still apply
+an unparseable impl-report.json halts
+an unterminated block comment before the marker is NOT an up section
+an up section plus a -- DOWN marker passes
+and does not claim the down section is missing
+and is reported by name
+and it reports the up-section failure
+and it reports the up-section failure
+and it reports the up-section failure
+and its own remedy, which is not the commented-out one
+and names the correct work it refuses
+and names the label that went unanswered
+and names the remedy for the block form specifically
+and names the remedy in the convention the report already uses
+and reports that file
+and says token overlap was deliberately not consulted
+and the clean down region is not reported
+and the halt quotes the criterion
+and the message quotes the DEFAULT marker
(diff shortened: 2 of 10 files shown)


nsmedia-io:
**Verified on both platforms. The gap is closed and the totals now agree exactly.**

`tests` run [32518574743](https://github.com/nsmedia-io/agent-pipeline/actions/runs/32518574743) on `ubuntu-latest`, against a local run of the same commit on macOS:

| suite | before (CI) | before (local) | after — both |
|---|---|---|---|
| `test-mis-tier-tripwire.sh` | 70 | 78 | **79** |
| `test-panel-composition-fail-direction.sh` | 141 | 165 | **165** |

```
34 suites, 2266 assertions, failed=0 — on ubuntu-latest
34 suites, 2266 assertions, failed=0 — on macOS
per-suite disagreements: NONE
```

Every suite reports the same number on both platforms. The runner installed zsh, and both suites recorded `` ok CAPABILITY `zsh` is present `` under `PIPELINE_TESTS_REQUIRE_CAPABILITIES=1` — so the `[zsh]` columns that are the #17 veto's regression test now run where the gate is enforced, and an absent optional tool would have reddened the build by name instead of shrinking the suite in silence.

(The mis-tier suite gains one over the old macOS 78: the self-equal `"not-run" "not-run"` line is gone and the capability declaration replaces it. The panel suite's 165 is unchanged for the same reason — one line removed, one added.)

`test-issue17-integration.sh`'s AC41(c) fresh-checkout case, the known flake in #25, passed on both. Nothing to capture there.

Issue #63, opened 2026-08-21
The 24-hour in-flight ceiling is pinned by no test at either side of its boundary, and the green window is roughly 1h to 25h


nsmedia-io:
Found by QA while authoring #43's Phase 3a contract, deliberately kept as that battery's expected survivor rather than closed, and re-measured by QA at Phase 4. Written here per the deferral rule before #43 merges. Not #43's subject, and that spec was at its acceptance-criteria ceiling.

## The gap

`gate-phase-entry.mjs` decides a run is in flight when it was updated less than `IN_FLIGHT_MS` ago and carries no final verdict. That constant governs whether the guard evaluates a record at all: outside the window it abstains from everything.

**No test pins either side of that boundary.** The suite's only fixtures are a fresh record and one aged 25 hours, so the constant can move a long way in both directions with everything green.

Measured at `2505a3d`, one mutation at a time, three gate suites (329 + 37 + 59 = 425):

| `IN_FLIGHT_MS` | result |
|---|---|
| `2 * 60*60*1000` (2h) | **329/0, fully green** |
| `24.9 * 60*60*1000` | **329/0, fully green** |
| `0.5 * 60*60*1000` | red, many |
| `26 * 60*60*1000` | red, 1 (the 25-hour fixture cell) |

So the green window is roughly **(1h, 25h)**, bounded only by the two fixture ages and not by any assertion about the constant itself.

QA's first estimate, read off the code at Phase 3a, was "an hour in either direction". Measuring it at Phase 4 showed that understated the looseness by an order of magnitude. Recording both, because the difference between a read figure and a measured one is the point.

## Why it matters

A silent change to 2 hours would make the guard refuse to evaluate any run left alone over a long lunch, and the entire suite would stay green. Direction is toward abstention, so the guard claims less rather than more, which is why this is a gap rather than a hazard. But the constant is load-bearing for every other assertion in that file: everything downstream only runs for a record inside the window.

## Why it could not be closed in #43

`git grep -n iso_hours_ago plugins/pipeline/tests/` returns only `test-gate-phase-entry.sh` — the helper at `:36` and the two fixture ages at `:38` and `:39`. No other suite uses it.

The closing test is a **boundary pair computed from the constant**, so it cannot drift when the constant legitimately changes. That needs the constant, or a `hoursInFlight()` seam, to be reachable from the suite. It is module-private today, which is exactly why #43's contract could not reach it.

## Shape of a fix

[...]

Issue #43, opened 2026-08-20
Undatable status.json records: the in-flight branch is unexercised, and inverting it wedges a project permanently


nsmedia-io:
Found by QA in #39 Phase 4 (2026-08-19). Mutation **M15b** (`!Number.isFinite(updated)` → `return true`) and **M15d** both SURVIVED every suite.

M15d surviving independently proves the stronger point: **no fixture in any suite ever omits `updated_at`**. `mk_status()` always interpolates `${MK_UPDATED:-$FRESH_ISO}`, so omitting it is structurally impossible through the suite's own builder. AC15's four cells are 25h-old, 1h-old, captured-stale, and final-verdict. None is undatable.

**Why direction matters.** If that branch inverted, an undatable record at a guarded phase would refuse **every turn in the project, permanently, with no clearance route** — the exact wedge the in-flight predicate exists to prevent. The test claims coverage it does not have.

Missing cells, named by QA:
- `AC15(e)`: `updated_at` absent → `not-applicable`
- `AC15(f)`: `updated_at: "not-a-date"` and `null` → `not-applicable`

Each paired with the fresh-`updated_at` control returning `refused`, so the cell discriminates rather than passing over an empty population.

nsmedia-io:
Phase 2 round 1 returned REQUEST_CHANGES (SecOps; DBA approve-with-notes, DevOps approve). Spec revised.

The substantive correction: the fix does not *narrow* the `1-ba` map check to the architectural tier, it **retires it at every tier** on the path `pipeline.md` mandates. A row gated on a resolved tier is unreachable at the only visit where `map.json` can still be missing; it survives only for rework re-entry, when the map necessarily already exists. SecOps measured 1-of-15 guarded rows catching a mapless architectural run today, 0-of-15 after, with a non-zero control on the same patched build. `gate-phase-entry.mjs:86` is the only executable requirer of `map.json` in the repo.

The direction stands (all three reviewers agree): today's row refuses trivial, standard and architectural identically at `1-ba`, so what is traded away is an *undiscriminating* refusal, in exchange for removing the pressure that trains operators onto the guard's own widest disarm. The lost coverage is re-sited in **#61**, filed before this merges, and the guard header will cite `#61` as a greppable token so the follow-up cannot be dropped quietly.

[...]

PR #70, opened 2026-08-21
fix(gate): re-site the map.json requirement to the 2-review row (#61)

### LINKED ISSUE #61: The architectural map requirement is sited at 1-ba, where the tier is never known
Split out of #43 / #45 at Phase 2 by SecOps, so that the loosening #43 ships does not leave a dead row behind unrecorded. **In scope for this issue, out of scope for #43's PR.**

## The problem

`plugins/pipeline/scripts/gate-phase-entry.mjs:86` (at `origin/main` `2ec6dd7`):

```js
"1-ba": { file: "map.json", tokens: ["0.5"], tiers: ["architectural"] },
```

This is the ONLY executable requirer of `map.json` anywhere in `scripts/` or `hooks/`:

```
$ git grep -n 'map\.json' origin/main -- plugins/pipeline/scripts plugins/pipeline/hooks
gate-phase-entry.mjs:86:  "1-ba": { file: "map.json", tokens: ["0.5"], tiers: ["architectural"] },
gate-phase-entry.mjs:99:  "0.5-map-complete": { file: "map.json", tokens: ["0.5"] },
```

(line 99 only guards the phase whose own name says the map was produced, so it cannot catch a run that never mapped.)

The row is gated on `tiers: ["architectural"]` at a phase where `commands/pipeline.md:103` guarantees the tier is **not yet known**: the `1-ba` checkpoint is committed BEFORE BA is dispatched, and `risk_tier` is BA's output. Today the row fires on every run at `1-ba` regardless of tier, because `normalizeTier` resolves an absent tier to the strictest one — an undiscriminating refusal of the first turn of every non-architectural run, which is the defect #43 fixes.

#43's fix makes the row not apply when the tier is undetermined. That is the right call (three reviewers agree), but it means the row is **retired on the mandated path at every tier**. It survives only for second-and-later visits to `1-ba` (BA rework re-entry, `pipeline.md:343`) — exactly when `map.json` necessarily already exists and the check can no longer catch anything.

## Measurement (SecOps, at `origin/main` `2ec6dd7`, pristine extract)

An architectural run that never ran Phase 0.5 (no `map.json` on disk, no `0.5` token in `events[]`) walked through all nine checkpoints it passes, `1-ba` through `5-archive`:

- **Today:** exactly **1 of 15** guarded rows refuses it (`1-ba`, rc 2).
- **After #43's patch:** **0 of 15** — every cell decides `granted` or `not-applicable`, rc 0.
- **Non-zero control on the same patched build:** `1-ba` with a RESOLVED architectural tier and no map still refuses (rc 2), so the harness was alive.

## Why it was not fixed inside #43

Re-siting the requirement is a behavioural change at a different phase with its own blast radius, and #43's diff is already a halting-control change plus two test-cell families. Folding it in would widen that diff past its size tripwire and mix two rulings in one review.

## Cheapest candidate fix (named so this issue is actionable, not prescriptive)

Make `1-ba-complete` a `byTier` row, since at `1-ba-complete` BA has returned and `pipeline.md` has copied the tier into `status.json`:

```js
"1-ba-complete": { byTier: {
  architectural: { file: "map.json", tokens: ["0.5"] },
  standard:      { file: "spec.json", tokens: ["1"] },
  trivial:       { file: "spec.json", tokens: ["1"] },
} },
```

Whoever takes this must check what `1-ba-complete` requires today before assuming the row is free, and must decide what happens to the now-dead `tiers` column on `1-ba` (delete it, or keep it for the re-entry case with a comment saying that is all it does).

## Acceptance sketch

- An architectural run with no `map.json` and no `0.5` event is refused by at least one row on a full nine-checkpoint walk, with a non-zero control (a run that DID map is not refused).
- A standard-tier and a trivial-tier run with no `map.json` are NOT refused at any checkpoint.
- The mutation that deletes the re-sited requirement reddens a named assertion.

Blocked on: #43 merging first (it owns `appliesAtTier`).

nsmedia-io:
Closes #61

## The defect

`gate-phase-entry.mjs` is the only executable requirer of `map.json` outside the phase that produces it, and it sited that requirement on the `1-ba` row behind `tiers: ["architectural"]`. `1-ba` is checkpointed BEFORE BA is dispatched and `risk_tier` is BA's output, so since #43 landed (a tier-restricted row abstains when the tier is undetermined) that row could not police a first visit at any tier. It looked live and was dead.

Measured at the merge-base: an architectural run with no `map.json` on disk and no `0.5` token in `events[]`, parked at `2-review`, was **granted, rc 0**. That is the run the requirement exists to catch.

## The change

The obligation moves to the `2-review` row, which pipeline.md checkpoints on the architectural route only, and which runs actually occupy (3 distinct records; the previously proposed `1-ba-complete` has never been a persisted `current_phase` in any commit of any ref).

**EXTEND, not swap.** The row gains ONE optional `also: { file: "map.json", tokens: ["0.5"] }` key beside its existing `spec.json` + `ba-approved` check. Swapping would have deleted the ba-approved gate from the entire architectural route: `2-constraints` carries that check but is standard-only by routing, and `1-ba-complete` is never occupied. Measured under a swap prototype, an architectural record with an unapproved `spec.json` goes rc 2 -> rc 0 and 12 existing assertions redden.

**No tier restriction of any kind** on the re-sited row. The phase NAME is the tier evidence, and a better one than the record's own field: `risk_tier` is optional in `status.schema.json` and no script writes it, while `current_phase` is written by the orchestrator following pipeline.md's routing. The `tiers` key stays on `1-ba` for the rework re-entry, so the re-derivation command pipeline.md publishes still returns exactly one hit.

**Primary first, with a short circuit.** The decision is the AND either way, so the ba-approved obligation is un-swappable by construction. What the ordering buys is that a doubly-failing row names ONE file, which keeps the single-route refusal shape `route1()` depends on — an address bought by a wrong-repair mutation that once survived a whole-stderr assertion.

`satisfyingTokens` returns the UNION over both halves. That is the only surface through which the drift suite enumerates what the guard accepts, so a token honoured by the decision path but not reported there would be invisible to the one check that exists to catch strays. The union is a REPORTING surface and is never fed back into `eventsSatisfy`: each half matches its OWN tokens, or a `1-ba` event would stand in for the map.

## The arity ceiling, made loud

`also` expresses exactly TWO requirements. A third, written as a nested `also.also`, is measured to be BOTH invisible to the token walk (drift suite 37/0, blind) AND **inert on the decision path** (rc 0 with its file absent) — a written requirement silently not enforced, the guard claiming more than it knows.

`rowShapes()` plus `#61-S1..S4` convert that permissive silence into a build failure: every row's key set must be a subset of `{file,tokens,content,tiers,byTier,also}` and every `also`'s a subset of `{file,tokens,content}`, asserted as an outcome over the whole live table with the permitted sets stated in the TEST rather than by the table itself. Recorded proof that the control bites: **RED on the planted violation (447/4), GREEN on the real state (451/0).**

## Version skew and back-out (R11)

- **Forward skew, intended.** An architectural run already parked at `2-review` without a `map.json` starts being REFUSED at its next turn boundary once an adopting project upgrades past this change. The refusal is actionable and offers three reachable routes: re-run the producing phase, record a SKIPPED event with a non-empty note, or conclude the run.
- **A clean upgrade is not proof the hole is closed for a given run.** A run already PAST `2-review` is never re-evaluated against this row.
- **Back-out is clean.** The rule table is CODE, not persisted state. No committed record's meaning changes: every historical record at `2-review` is excluded from evaluation by `inFlight()` (24h + final_verdict) regardless of the row. The down path is a plain revert of the `also` key plus the `prerequisiteSatisfied`/`decideForDir` indirection — no data to unwind, no schema cache, no persisted row shape.
- **Sequencing.** #61 and #63 must merge SEQUENTIALLY, never concurrently: both edit `gate-phase-entry.mjs` and add fixture families to `test-gate-phase-entry.sh`, and the `GUARDED_ROWS`/AC1/AC2 lockstep is exactly the bash-array literal a concurrent second editor leaves silently stale without a merge conflict.

## Verification

`bash plugins/pipeline/tests/run.sh` — 34 suites, **2493 assertions, 0 failed, 0 skipped**. The harness has no skip path by construction (#47), so no suite can pass by abstaining.

- Gate suite `414/33 -> 451/0` (the 33 are QA's Phase-3a contract; the 4 added are graft G2)
- Drift suite `37/0`, zero drift-suite edits
- `#43-S1/S2/S3` re-measured against this diff rather than inherited: green

Mutation battery, all under commit-first hygiene (literal replacement, `count == 1` asserted, changed line printed and confirmed to be the line meant, restored with `git checkout --`, `git status --porcelain` empty after each):

| id | mutation | expected | result |
|---|---|---|---|
| AC11-DEL | delete the `also` key | RED | 417/34, contains the named label `#61-C1` |
| A1 | stray token in `also.tokens` | RED | drift 36/1 |
| A2 | non-zero control: same stray in PRIMARY tokens | RED | drift 36/1 |
| A3 | `also` with a stray on a byTier cell | RED | drift 36/1 |
| **A4** | stray in a nested `also.also` | **SURVIVE** | drift 37/0 — the documented arity ceiling |
| G2-BITES | the same `also.also` shape | RED then GREEN | 447/4 planted, 451/0 restored |
| AC7-DEL | delete `tiers` from the `1-ba` row | RED | `#61-G1`/`#61-G2` |
| P-a | re-introduce a retired-at-every-tier sentence | RED | `#61-P2` alone |
| P-b | insert the paired negative clause | RED | `#61-P1` alone |
| P-c | point the siting sentence at another row | RED | `#43-K3` alone |
| P-d | change EARLIER to PRIOR inside ANCHOR_43 | RED | `#43-K2`, `#61-P3` |
| **MUT-S** | drop `Array.isArray` from the tooling fail-open guard | **SURVIVE** | 451/0, 37/0, 59/0 |

Two documented survivors, so "all red" cannot be mistaken for coverage. P-b and P-c redden **disjoint** single assertions, which is what proves neither half of the prose pair blinds the other.

Three spec-vs-implementation deltas QA measured rather than assumed are recorded in `impl-report.json` so the panel does not re-derive them — most notably that AC12's named mutation is a **no-op** at the spelling the spec placed it (` architectural` already resolves to the strictest row and `tierDetermined` stays false either way), and the discriminating spelling is ` trivial` at `3-impl`.

### DIFF
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -330,23 +330,44 @@ export function satisfyingTokens(phase, tier) {
 }
 
 /**
- * Every key set in the rule table above -- one entry per row, per `byTier` cell and per `also`
- * sub-row, tagged with the kind of position it occupies. The PERMITTED key sets deliberately
- * live in the test and not here: a table that published its own permitted keys would be graded
- * against its own opinion, and the thing worth catching is a key nobody taught the walk to read.
+ * Every key set in the rule table above, one entry per position, tagged with WHICH POSITION it
+ * occupies. The PERMITTED key sets deliberately live in the test and not here: a table that
+ * published its own permitted keys would be graded against its own opinion, and the thing worth
+ * catching is a key nobody taught the walk to read.
+ *
+ * FOUR KINDS, NOT TWO, because the same key name means different things depending on where it is
+ * written, and a check total over NAMES is not total over POSITIONS. Each kind is named for the
+ * set of readers that reach it:
+ *   - `row`        a plain top-level row. Every reader reaches it.
+ *   - `dispatcher` a top-level row carrying `byTier`. `rowFor` returns the CELL, so `file`,
+ *                  `tokens`, `content` and `also` written here are never read -- and a `file`
+ *                  written here is a requirement the guard grants without checking.
+ *   - `cell`       a `byTier` cell. `appliesAtTier` reads `tiers` off the RAW top-level row, so
+ *                  a `tiers` written on a cell is never read.
+ *   - `also`       a second-requirement sub-row. Its arity ceiling is two, so a nested `also`
+ *                  here is the third requirement nothing enforces.
+ * Every inert position named above was driven through the CLI rather than reasoned about, and
+ * `tiers` on a DISPATCHER is the one key that survives that test: it IS read, so it stays in the
+ * dispatcher's permitted set. Excluding a key the guard honours is the same error pointed the
+ * other way.
  */
 export function rowShapes() {
   const out = [];
+  // The kind is a property of the POSITION, not of the contents: a `byTier` nested inside a cell
+  // is tagged `cell` and its `byTier` key is a stray, which is what it deserves -- it disarms the
+  // row it is written on. Tagging by contents would re-admit it as another dispatcher.
   const walk = (at, kind, obj) => {
     out.push({ path: at, kind, keys: Object.keys(obj).sort() });
     if (obj.byTier) {
       for (const [cellTier, cell] of Object.entries(obj.byTier)) {
-        walk(`${at}.byTier.${cellTier}`, "row", cell);
+        walk(`${at}.byTier.${cellTier}`, "cell", cell);
       }
     }
     if (obj.also) walk(`${at}.also`, "also", obj.also);
   };
-  for (const [phase, raw] of Object.entries(PREREQUISITES)) walk(phase, "row", raw);
+  for (const [phase, raw] of Object.entries(PREREQUISITES)) {
+    walk(phase, raw.byTier ? "dispatcher" : "row", raw);
+  }
   return out;
 }
 
@@ -424,10 +445,17 @@ function checkOne(issueDir, subRow, events) {
  * failing row names ONE file, which is the single-route shape the refusal template and its test
  * both depend on. The cost is that an operator who fails both halves at once returns for a
  * second refusal after fixing the first; both refusals are correct and both are actionable.
+ *
+ * A VACUOUS PRIMARY IS A VALUE, NEVER AN EARLY RETURN. A row with no `file` owes nothing on its
+ * primary half, and it is tempting to answer `ok` and stop -- but stopping there skips `also`
+ * entirely, so a second requirement written on such a row would be reported by
+ * `satisfyingTokens` and never enforced: the guard claiming more than it knows, which is the
+ * one failure this row's own arity ceiling exists to prevent. No shipped row is shaped that way,
+ * and the fall-through is what keeps that a fact about the table rather than a fact about the
+ * table's current contents.
  */
 function prerequisiteSatisfied(issueDir, row, events) {
-  if (!row.file) return { ok: true, artifact: "none", file: null, content: null };
-  const primary = checkOne(issueDir, row, events);
+  const primary = row.file ? checkOne(issueDir, row, events) : { ok: true, artifact: "none" };
   if (!primary.ok) return { ...primary, file: row.file, content: row.content || null };
   if (row.also) {
     const secondary = checkOne(issueDir, row.also, events);
@@ -435,7 +463,12 @@ function prerequisiteSatisfied(issueDir, row, events) {
       return { ...secondary, file: row.also.file, content: row.also.content || null };
     }
   }
-  return { ok: true, artifact: primary.artifact, file: row.file, content: row.content || null };
+  return {
+    ok: true,
+    artifact: primary.artifact,
+    file: row.file || null,
+    content: row.content || null,
+  };
 }
 
 /**
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -83,12 +83,33 @@ import { activeIssueDir } from "./validate-pipeline-artifact.mjs";
  * `content` marks a row whose prerequisite is not satisfied by mere presence. Such a row is
  * strictly weaker through events[] than through the file, because an event attests DISPATCH,
  * never APPROVAL, and that asymmetry is stated rather than defended against.
+ *
+ * `also` is a SECOND requirement on the same row, carrying its own file, its own token set and
+ * its own optional `content`. It is checked only after the primary one passes, so a refusal
+ * never names two files at once and the message keeps its single-route shape. THE ARITY IS TWO,
+ * and the ceiling is defended rather than assumed: a third requirement written as a nested
+ * `also.also` would be both invisible to `satisfyingTokens` and inert on the decision path -- a
+ * written requirement silently not enforced, the guard claiming more than it knows -- so the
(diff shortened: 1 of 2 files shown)


nsmedia-io:

|---|---|---|
| BA | APPROVE_WITH_NOTES | none outstanding (its REQUEST_CHANGES on the deferral ledger was discharged and re-verified at the destination) |
| DBA | APPROVE_WITH_NOTES | none outstanding (both majors implemented) |
| DevOps | APPROVE_WITH_NOTES | none |
| SecOps | APPROVE_WITH_NOTES | none; 0 critical / 0 high / 0 medium |
| Dev | APPROVE_WITH_NOTES | none outstanding (its own finding fixed in round 1) |
| QA | APPROVE_WITH_NOTES | none outstanding (its REQUEST_CHANGES closed by `#61-S8/S9/S10`) |

**Final verdict: APPROVE_WITH_NOTES.** Two delta re-review rounds; QA and SecOps were re-dispatched unconditionally on each.

### What the panel changed about the change

Three findings were fixed rather than deferred, and two of them were defects the remediation itself introduced or widened:

1. **`prerequisiteSatisfied`'s vacuous-primary early return skipped `also`** (found by planting it, not by inspection). Fixed; `#61-V1..V5` drive the real CLI.
2. **The shape walk graded key names, not positions.** Two silent-grant survivors were demonstrated. `rowShapes()` now grades four positions; `#61-S5/S6/S7`.
3. **A key cannot be graded without its sibling.** `{file: null, tokens: []}` and `{file: null, tokens: ["0.5"]}` are the same key set, so no subset check separates them. `#61-S8/S9/S10`.

Finding 3 mattered more than it first looked. Fix 1 widened a **silent fail-open**: `main()` collects decisions in a loop and prints them only afterwards, and the fail-open catch wraps the whole call, so one malformed row discards every decision already computed — including a *different* issue dir's genuine `rc 2` refusal. Since `hooks/stop.sh` branches only on `rc 2` and discards stdout, that reads as a pass. Measured across three refs on a two-dir fixture; `#61-S9` makes the only table shape that reaches the throw a build failure.

### Verification

- Suite **2530 passed / 0 failed** across 34 suites, **0 skip lines, 0 cached lines**. Arithmetic control: 2527 + 3 new assertions = 2530.
- Nothing weakened: label-set `comm -23` shows the removed-id set **empty** across every round; one assertion (`#43-K3`) moved and was re-anchored with a paired negative and its own expiry sentence, per its own instructions.
- Every new assertion was watched to fail before it passed, with a non-zero control, and each round carries a **documented expected survivor** so an all-red battery cannot be mistaken for coverage.
- CI green on the reviewed head (`6ad1b13`), which is the PR head.

### Deferred, and written down

The deferral ledger is in two comments on #61 (not in the artifacts, which are gitignored and vanish with the worktree). It records the rejected `1-ba-complete` siting with its census, two `pipeline.md` prose handoffs owned by another lane, the unpinned fail-open catch, two same-class shapes one key over, and a validator gap found at merge time: **`validate-pipeline-artifact.mjs` does not resolve `$ref`**, so all 7 `$ref` constraints in `peer-review.schema.json` are documentation rather than guards.

Not on panel: Design and Art Director (no frontend surface, no visual contract).

Issue #74, opened 2026-08-22
The in-flight window is spelled independently in five places, one over a different column, and the abstention it governs is invisible in production


nsmedia-io:
Deferred out of #63 (Phase 2 review: DBA, SecOps, DevOps). #63 exports `IN_FLIGHT_MS` from `gate-phase-entry.mjs` and pins its boundary with a derived pair. It does **not** retune the value and it does **not** touch any file below — every one of these lives outside #63's lane. This issue is the tracked home for the class, filed before #63 merges so its corrected header has a real number to cite.

The class: **the 24-hour in-flight window is spelled independently in five places, one of which reads a different column than the guard does, one of which claims to share the guard's definition and does not, and none of which can witness the decision the window actually governs.**

## 1. `pipeline-status.mjs:156` holds a second, independent copy

```js
// plugins/pipeline/scripts/pipeline-status.mjs:156
return age > 24 * 60 * 60 * 1000 && !r.final_verdict;
```

That is the exact complement of the guard's predicate:

```js
// plugins/pipeline/scripts/gate-phase-entry.mjs:499
return now - updated <= IN_FLIGHT_MS;
```

**Measured: there is no shared symbol.** `grep -rn 'IN_FLIGHT_MS' plugins/ commands/ agents/` returns only `gate-phase-entry.mjs:163` (the declaration) and `:499` (its only use), plus one hit in `knowledge/issue-archive/43.json`. `gate-phase-entry.mjs` imports nothing from `pipeline-status.mjs`. The two literals can diverge while both modules compile and every suite stays green.

The guard asserted the opposite in two places. #63 corrects both:

- `:26-27` — *"reusing pipeline-status.mjs's own 24h / no-final-verdict definition rather than inventing a second one."* Flatly false.
- `:490-491` — *"the predicate pipeline-status.mjs already uses to call a run 'possibly stuck', inverted."* True **today**, and false the moment either literal moves, with nothing to say so.

**Ask here:** unify the two modules on one exported symbol. #63 deliberately does not, because `pipeline-status.mjs` is not its lane and because a unification is a behaviour-adjacent change to two consumers rather than a testing change to one.

## 2. `session-start.sh:164` reads a different column — a grain mismatch, not a third spelling

```bash
# plugins/pipeline/hooks/session-start.sh:164
[[ -n "$(find "$dir/status.json" -mtime -1 2>/dev/null)" ]] || continue
```

[...]

Issue #66, opened 2026-08-21
The SubagentStop artifact validator is inert in every plugin-installed project: AGENT_RULES is keyed by bare role names and a namespaced agent_type returns zero failures for every rule


nsmedia-io:
## The defect

`plugins/pipeline/scripts/validate-pipeline-artifact.mjs:93` declares `AGENT_RULES` keyed by **bare role names** (`ba`, `dba`, `devops`, `secops`, `qa`, `dev`, `librarian`, ...). `checkArtifacts` resolves the stopping agent against that table and returns early on a miss:

```js
export function checkArtifacts(agentType, input, now = Date.now(), rootsOverride) {
  const rules = AGENT_RULES[String(agentType || "").toLowerCase()];   // :654
  const failures = [];
  if (!rules) return { failures };                                     // :656
```

When the pipeline runs as an **installed plugin**, the agent type is namespaced (`pipeline:secops`). It matches no key, and the function returns **zero failures for every agent, every artifact, every rule** — before `pipelineDirs`, before `activeIssueDir`, before the `RECENT_MS` recency test. The SubagentStop artifact validator is therefore **inert in every plugin-installed project**, which is the shipping product's default deployment.

This is a cross-project defect. It is filed rather than fixed inside the run that found it (see Scope, below).

## Evidence

Everything below is re-derivable by a reader with no access to this run.

### 1. The four-cell fixture, against the shipped hook

Fixture: `knowledge/issue-archive/43.json` `.review.secops` (an archived SecOps shard with **7 known missing `description` fields**), written to `<scratch>/.pipeline/43/review.secops.json` with a fresh `status.json`. One fixture, one scratch project root, four values of `agent_type`:

| `agent_type` | outcome |
|---|---|
| `secops` | `decision:block`, naming all 7 errors — **positive control** |
| `SecOps` | `decision:block` — the `.toLowerCase()` normalisation works |
| `pipeline:secops` | **SILENT**, empty stdout, exit 0 |
| `agent-pipeline:secops` | **SILENT**, empty stdout, exit 0 |

Re-run it:

```sh
printf '{"agent_type":"pipeline:secops","cwd":"<scratch>"}' \
  | CLAUDE_PROJECT_DIR=<scratch> CLAUDE_PLUGIN_ROOT=<repo>/plugins/pipeline \
    bash plugins/pipeline/hooks/subagent-stop.sh
```

Swap `pipeline:secops` for `secops` to get the non-zero control back.

[...]

PR #77, opened 2026-08-22
fix(gate): pin the in-flight boundary to the constant the guard actually uses (#63)

### LINKED ISSUE #63: The 24-hour in-flight ceiling is pinned by no test at either side of its boundary, and the green window is roughly 1h to 25h

nsmedia-io:
Closes #63

## What

Two commits, two files, no behaviour change.

**`c95537a` (QA, Phase 3a)** — the failing behavioral contract in `plugins/pipeline/tests/test-gate-phase-entry.sh`. The suite went 462 → 492 passing assertions.

**`56e8671` (Dev, Phase 3b)** — `plugins/pipeline/scripts/gate-phase-entry.mjs`, +13/−5:

- `const IN_FLIGHT_MS` → `export const IN_FLIGHT_MS`. Value byte-identical (`24 * 60 * 60 * 1000`), `inFlight` and its single call site untouched. `inFlight` itself stays private: a second parallel path to the same answer is precisely what the declared `<=`/`<` survivor's justification rests on not existing.
- The header claim that the guard reuses `pipeline-status.mjs`'s window "rather than inventing a second one" is deleted. It was flatly false — no symbol is shared and both modules hold their own literal.
- `inFlight`'s doc comment loses "the predicate pipeline-status.mjs already uses to call a run 'possibly stuck', inverted" (true today, silently false the moment either literal moves) and gains a drift-risk paragraph sited at the predicate, citing #74 and naming `pipeline-status.mjs:156` as the independent copy. It also states the precise divergence: the two are inverses on the age term **alone**, since an undatable record is neither stuck there nor in flight here.

Before this, the boundary the guard enforces was untestable from outside without re-spelling `24 * 3600e3` in the suite — which is what both former sites did, judging fixtures against a ceiling they did not own.

## Verification

| | result |
|---|---|
| `bash plugins/pipeline/tests/run.sh` | 34 suites, **2560 passed, 0 failed**, no self-skips, no cached counts |
| entry / drift / hook | 492/0, 37/0, 59/0 (hook byte-identical to base) |
| baseline red at `c95537a` | `passed=0 failed=1` — `#63-R1`, then the suite HALTS by design (`finish || exit 1`) rather than reporting vacuous greens on a broken read |
| staging control (export only, prose reverted) | `passed=489 failed=3`, reds exactly `#63-H1`, `#63-H2`, `#63-H3` |
| drift immunity, 2h rewrite | 492/0 |
| drift immunity, 48h rewrite | 490/2 — reds exactly `#63-P2` and `(a) 25h old…`, whose premises are genuinely false at a 48h ceiling. The boundary pair and its control stayed green at both bases. |
| label-set containment (entry, drift) | 0 base labels lost, +30 / +0 added |

### Mutation battery

Isolated detached worktree, one mutation at a time, each applied by unique literal replacement (`count == 1` asserted, changed line printed) and restored with `git checkout --`.

| | mutation | expected | result |
|---|---|---|---|
| M1 | guard `<=` → `<` | **survive** | SURVIVED (entry 492/0, drift 37/0) |
| M2 | probe43 `>` → `>=` | **survive** | SURVIVED (492/0) |
| M3 | guard `<=` → `>=` | redden | CAUGHT — 261/231, incl. `#63-B1/B2/B3/C2/C3` |
| M4 | age term deleted (`return true`) | redden | CAUGHT — 486/6, incl. `#63-B2/B3/C2/C3` |

Two survivors are **declared and expected**, each documented in the suite with its mechanism, residual and expiry: both spellings differ only when the age is exactly the ceiling to the millisecond, and the guard reads its own `Date.now()` in a child process started after the fixture was written, so no CLI-driven fixture lands there. The residual is one millisecond in the direction of abstention — a silence, not a falsehood. A battery in which everything reddens cannot tell coverage from a rubber stamp.

Mutating the constant's **value** deliberately does not redden the pair (the 2h/48h runs above), which is why it is inadmissible as coverage evidence.

## Phase 4 fix round (commit `5bdc9bf`)

The panel returned **SecOps REQUEST_CHANGES** on one blocker plus a DBA major. Both are fixed here, with two related notes.

**BLOCKER — the round-1 corrected prose was ALSO false, in the same class as the claim it replaced.** It said the two predicates *"are inverses on the age term ALONE: an undatable record is neither stuck there nor in flight here."* The second half is false. `pipeline-status.mjs` builds its row as `updated_at: status?.updated_at ?? mtime`, so an **absent or null** `updated_at` is filled in from the status.json **file mtime** before its comparison. Measured on one record — `.pipeline/91/status.json = {"current_phase":"3-impl","events":[]}`, mtime touched to 2026-01-01:

```
pipeline-status.mjs  ->  ## Possibly stuck (over 24h, no final verdict)   - #91 at `3-impl`
gate-phase-entry.mjs ->  {"decision":"not-applicable","reason":"... is not in flight ..."}
```

Both fire on one record, which "inverses" says cannot happen. Non-zero controls on both axes: `updated_at` 1h ago → not-stuck / refused; 25h ago → STUCK / not-applicable. Why it mattered beyond pedantry: the comment's job is to send a future maintainer to #74, whose ask is to unify the two modules — and that sentence told them the datability paths already agree and need no reconciliation.

**Refinement measured beyond the blocker report:** `??` fires only on **absent** and **null**, so an **unparseable string** (`"not-a-date"`) with an old mtime is *not* stuck there and *not* in flight here. It is the one undatable spelling on which the two modules still agree, and the shipped prose now says exactly that rather than generalising over "undatable".

**Correction to a related QA finding, caught before it shipped.** QA reported that with `updated_at: 12345`, `Date.parse` reads year 12345 (permanently in flight here) while `new Date(12345)` reads epoch 12.345s, so pipeline-status.mjs calls the same record possibly-stuck — "both fire". **The second half is false.** pipeline-status.mjs does not classify a numeric `updated_at` at all; it **throws first**: `TypeError: (r.updated_at ?? "").slice is not a function` at `pipeline-status.mjs:130` in `renderMarkdown`, `rc=1`. The `stuck` filter is at `:153`, inside `renderMarkdown` and *after* the throwing line, so it is never reached; in `--json` mode `renderMarkdown` is never called at all. Shipping QA's sentence verbatim would have put a second measured-false claim into the very comment this round exists to correct. The divergence **is** recorded in the shipped prose, in the form actually measured.

**MAJOR (DBA) — `#63-P2`'s label was false.** It claimed the P1/P2 interval was *"the ONLY remaining bound on the VALUE of `IN_FLIGHT_MS` after #63 … because every other new cell is DERIVED from the value."* Measured false: `#63-A2` pins the constant's exact source text, is neither derived nor value-independent, and is strictly tighter. Rewriting the shipped constant to 2h gives **492/1 with `#63-A2` as the sole red**. The label is corrected, and `#63-A2` gained the two notes its siblings carry — a do-not-delete, and a failure-direction sentence (direction is toward **false alarm**: a reformat to `86_400_000` reddens it for a non-defect reason). **Consequence: SecOps' round-1 concern that "after #63 the value 2h is STILL green" is no longer true and should not be re-litigated.**

**LOW (SecOps) — the invariance tripwire missed the last write-capable call.** `#63-C4`/`#63-C5` sat after the *first* of two `ac63_guard_copy` call sites; the second had no invariance cell after it, and that is the call whose residue an interrupted run actually leaves in the checkout. Added **`#63-C6`**, re-asserting the checksum after the last copy-driven cell.

**NIT — the prose did the thing its own test warns against.** Both drift comments cited the sibling literal as the absolute `pipeline-status.mjs:156`, twice in one file, with nothing pinning either — while `#63-H3`'s own siting check deliberately uses a band *relative* to `function inFlight(` because "an absolute range decays the first time anything above it moves." Both absolute citations are gone; `grep -c ':156'` on the guard is now **0**. The sibling is cited once, by symbol.

### Fix-round mutation battery

Fix committed **first**, so every restore is `git checkout --` from a committed state. Each mutation applied by literal replacement through a helper that **refuses unless the literal matches exactly once**; changed line printed from the file after the edit; restored from git; `git status --porcelain` verified after each. Guard mutations were driven against a **copy** via the `SCRIPTS_DIR` override, so nothing was written to the tracked guard except the two checkout-value rewrites, both restored and verified clean.

Redirected baseline `SCRIPTS_DIR=<copy>`: **493/0**. Unredirected: **493/0**.

| | mutation | expected | result |
|---|---|---|---|
| MUT-A | write the real residue (`const IN_FLIGHT_MS = 86400000;`) to the checkout after the **last** copy call | redden `#63-C6` | CAUGHT — 490/3, `#63-C6` red, **`#63-C4` and `#63-C5` both GREEN** (the blind spot) |
| MUT-B | same siting, semantically inert write (`printf '\n' >>`) | redden `#63-C6` | CAUGHT — **492/1, `#63-C6` the SOLE red** |
| MUT-C | shipped constant → 2h (checkout) | redden `#63-A2` only | CAUGHT — 492/1, `#63-A2` sole red, P1/P2 green |
| MUT-D | shipped constant → 48h (checkout) | redden `#63-P2` | CAUGHT — 490/3 (`#63-P2`, the `(a) 25h old` label, `#63-A2`) |
| MUT-D′ | same 48h rewrite, in a **copy** | redden the declared pair only | CAUGHT — 491/2, `#63-A2` **green** |
| MUT-E | strip both `#74` tokens from the band above `function inFlight(` | redden `#63-H3` | CAUGHT — 492/1, `#63-H3` sole red |

**MUT-B is the load-bearing one:** a single trailing newline written to the checkout after the last copy call is invisible to every other cell in the file, including `#63-C4` and `#63-C5`. That is the gap `#63-C6` closes.

**MUT-D′ reconciles a discrepancy rather than leaving it.** DBA reported the 48h rewrite as "490/2, no third red"; this round measured 490/3. Both are correct — `#63-A2` reads the **checkout**, so it is blind to a copy mutation and reddens only when the shipped file moves. The reading depends entirely on which file was mutated.

**Both declared survivors re-verified and still surviving:** `<=` → `<` in `inFlight` (493/0) and probe43's `>` → `>=` (493/0).

> **Harness disclosure.** The first attempt at survivor 2 was a **false green**. The replacement literal carried a trailing newline the file did not have, so it matched 0 times; the uniqueness guard **refused the edit** and the suite that ran was unmutated, reporting a meaningless 493/0. Re-done with the correct literal, verified applied by printing the changed line *and* asserting 0 remaining occurrences of the original spelling. Recorded because a battery that cannot tell "survived" from "never applied" is a rubber stamp.

## Delta re-review round (commit `a90dd19`)

QA and SecOps **independently found the same single blocker**: the paragraph I *added* in the previous round contained a third measured-false universal.

> ~~which agree on the ISO strings `status.schema.json` requires and diverge on anything else~~

**`new Date(string)` delegates to `Date.parse`**, so the two agree on *every* string, not only the ISO ones. Verified independently rather than relayed — 10 formats, **0 divergences**:

| input | `Date.parse` | `new Date().getTime()` | |
|---|---|---|---|
| `"2026-08-20T00:00:00Z"` | 1787184000000 | 1787184000000 | AGREE |
| `"March 3, 2020"` | 1583211600000 | 1583211600000 | AGREE |
| `"2020/03/03"` | 1583211600000 | 1583211600000 | AGREE |
| `"Tue, 03 Mar 2020 00:00:00 GMT"` | 1583193600000 | 1583193600000 | AGREE |
| `"garbage 99"` | 915166800000 | 915166800000 | AGREE |
| `"not-a-date"` / `""` | NaN | NaN | AGREE |
| `12345` | 327403400400000 | 12345 | **DIVERGE** |
| `0` / `false` / `true` / `null` / `1.5` | — | — | **DIVERGE** |

They part **only on values that are not strings**, where `Date.parse` coerces to string first and the constructor treats a non-string primitive as a *time value*. The clause also **contradicted the immediately preceding paragraph**, which correctly says an unparseable string is the one undatable spelling the two agree on — two lines below the words "Measured, not reasoned."

**One refinement I measured beyond the drafted fix:** `[]` and `{}` are non-strings that **agree** (both `NaN`), because the constructor's `ToPrimitive` yields a string that is then parsed. So "all non-strings diverge" would have been a *fourth* false universal. The shipped sentence states "part **only** on values that are NOT strings" as a **necessary** condition, not a sufficient one — that distinction is load-bearing.

**QA-N1 folded in, and tightened rather than kept.** "It throws on the number before its filter is reached" was precise only for markdown mode with one issue dir. I measured all four configurations:

| dirs | mode | result |
|---|---|---|
| 2+ | markdown **and** `--json` | throws in `listActive`'s sort comparator, `pipeline-status.mjs:106` (`localeCompare is not a function`), rc=1 |
| 1 | markdown | throws in `renderMarkdown`, `:130`, rc=1 |
| 1 | `--json` | **no throw**, rc=0 — and never runs the filter either |

So the shipped sentence now states the load-bearing property — **the filter is not reached on any path** — instead of naming one site.

### Corrected paragraph, as shipped

```
 * They also DATE the field differently -- `Date.parse` here, `new Date(...).getTime()` there --
 * and MEASURED, those two agree on every STRING, not only the ISO ones status.schema.json
 * requires: `new Date(string)` delegates to `Date.parse`, so "March 3, 2020", "2020/03/03" and
 * "not-a-date" all land identically in both. They part only on values that are NOT strings,
 * where `Date.parse` coerces to string first and the constructor does not. `updated_at: 12345`
 * reads here as the year 12345, so the record is permanently in flight, while
 * pipeline-status.mjs never classifies it at all: its filter is not reached on ANY path. With
 * two or more issue dirs it throws first in `listActive`'s sort comparator, in both markdown
 * and `--json`; with one dir it throws in `renderMarkdown`; and the sole non-throwing case, one
 * dir under `--json`, does not run the filter either. Nothing in this tree pins either spelling.
```

Applied uniquely (`count == 1`) with **zero occurrences of either original clause remaining**, verified by grep — that second check is what caught an earlier false green in this PR. `#63-H1`/`H2` at 0 hits, `#63-H3` still sited (band 504–523, `#74` at 519/520). Suite **493/0**; full runner **34 suites / 2561 / 0**.

### `MUT-A` call site — answering SecOps' non-blocking question

**I never repointed `AC63_COPY_GUARD`.** The helper `ac63_guard_copy` was not edited at all: its only assignment, `AC63_COPY_GUARD="$NEW_TMPDIR/scripts/gate-phase-entry.mjs"`, is untouched at `test-gate-phase-entry.sh:2499`.

`MUT-A` inserted **one new standalone statement immediately after the second and last `ac63_guard_copy` call** — the `no-export driver`, at what was line 2675 — writing directly to `$SCRIPTS_DIR/gate-phase-entry.mjs` via `node -e`. The first call site was not modified. That is exactly why `#63-C4`/`#63-C5` stayed green: **they had already run before the only write occurred.**

SecOps applied the criterion's *own named mutation* instead (repoint the helper at `$SCRIPTS_DIR`), which is **global** — it writes at the first call too, before C5's checksum — so C5 correctly reddens there. SecOps' reasoning was sound for the mutation it ran. Both readings are right for the mutation each applied, and they answer different questions: the global variant proves the class is caught *somewhere*; the targeted variant isolates *which cell* catches a write performed **after** C5 — the gap `#63-C6` exists to close. QA reproduced the targeted 490/3 with C4/C5 green.

## Round 4 — the enumeration is deleted (commit `4a62295`)

QA and SecOps **independently** found a fourth false claim in the same paragraph, and converged on the same root cause *and* the same fix. Two clauses I added in round 3:

> ~~With two or more issue dirs it throws first in `listActive`'s sort comparator, in both markdown and `--json`~~
> ~~the sole non-throwing case, one dir under `--json`~~

**Root cause.** The comparator is `(b.updated_at ?? "").localeCompare(a.updated_at ?? "")`. `localeCompare` **coerces its argument**, so a bad value in `a` is harmless — it throws only when the bad row is the **receiver `b`**. V8's insertion sort passes the element being inserted as `a`. The governing variable is therefore **where the offending row lands in readdir order**, not the dir count, which is exactly what my clause keyed on.

Reproduced before shipping the fix — two fixtures, identical readdir order `["10","63"]`, differing only in which dir carries the bad value:

| fixture | markdown | `--json` |
|---|---|---|
| bad row **first** | rc=1, `listActive` | rc=1, `listActive` |
| bad row **last** | rc=1, `renderMarkdown` | **rc=0, no throw** |

The second row alone falsifies both clauses. QA's tally across 20 cells: cmp=10 / `renderMarkdown`=5 / no-throw=5, where my sentence predicted 18/1/1 — **8 of 18 multi-dir cells contradict it**. SecOps ran n=1..1000.

**Why this is a deletion and not a fifth draft.** It is not a clean "not last" rule either: a mid-position bad row **survives** at n=5,10,22,23,24 and **throws** at n=3,32,64,100,1000, because binary insertion probes only an O(log n) prefix. Any characterisation of the throw site is a claim about a V8 sort internal — pinning it would be the fifth false claim, this time with a green checkmark behind it.

### Final paragraph, as shipped

```
 * They also DATE the field differently -- `Date.parse` here, `new Date(...).getTime()` there --
 * and MEASURED, those two agree on every STRING, not only the ISO ones status.schema.json
 * requires: `new Date(string)` delegates to `Date.parse`, so "March 3, 2020", "2020/03/03" and
 * "not-a-date" all land identically in both. They part only on values that are NOT strings,
 * where `Date.parse` coerces to string first and the constructor does not. `updated_at: 12345`
 * reads here as the year 12345, so the record is permanently in flight, while
 * pipeline-status.mjs never classifies it at all: its `stuck` filter is not reached on any
 * path. Markdown mode always throws before it -- in `listActive`'s sort comparator or in
 * `renderMarkdown`'s table loop, depending on where the offending row lands in readdir order --
 * and `--json` never calls `renderMarkdown` at all, so the filter does not run there even when
 * nothing throws. Nothing in this tree pins either spelling.
```

It makes **no claim about which site fires or how many dirs are needed** — which is precisely why it cannot be falsified the way the last four were. Its two remaining premises both hold: markdown always throws before the filter (structurally, `renderMarkdown`'s table loop at `:130` does `(r.updated_at ?? "").slice(0,19)` over every row and precedes the filter at `:153`), and `--json` never calls `renderMarkdown` at all.

Applied uniquely (`count == 1`) with **zero occurrences of all four original clause fragments remaining**, grep-verified. `#63-H1`/`H2` at 0 hits, `#63-H3` sited (band 505–524). Suite **493/0**; runner **34 suites / 2561 / 0**.

**Held under attack and deliberately not re-touched:** "agree on every STRING" (SecOps: 47-case corpus + **300,000 random strings**, zero disagreements, oracle control firing both ways; QA: 14/14, and it is *specified* behaviour, not merely empirical); "part only on values that are NOT strings" (both confirm the necessary-not-sufficient reading; QA measured 9 of 16 non-strings agree); "Nothing in this tree pins either spelling" (settled by mutation, both swaps survive with controls reddening in the same session); "year 12345 / permanently in flight".

## Known limits and what is left undone

`.pipeline/` is gitignored in this repo, so the implementation report does not survive the worktree. Everything load-bearing is therefore written here.

1. ~~**`AC2` is PARTIAL.**~~ **DISCHARGED — `AC2` is now PASS.** The hook suite's label-set capture, the one half missing at hand-off, has been captured **three times independently** — QA and BA in Phase 4, and Dev in this fix round — all agreeing: **58 unique labels at base, 58 at head, 0 lost, 0 added, 59/0 at both ends.** Non-zero control on the diff itself: the same `comm -23` against a deliberately truncated 3-label set reports 55 losses, so the zero is a measurement, not a broken pipeline. QA ruled explicitly that the byte-identity argument alone was **not** sufficient — byte-identity pins the label *strings* but not which cells *run*, and the hook suite is guard-sensitive — so PARTIAL was the honest grade at hand-off and is discharged by capture rather than by argument. Entry suite is now 362 → 393 (0 lost, 31 added, including `#63-C6`); drift 37 → 37.
2. ~~**The `<=` → `<` survivor on the hook suite is owed.**~~ **DISCHARGED.** Re-verified surviving; the full runner is green at 34 suites / **2561 passed / 0 failed**, hook 59/0 included.
3. **The third `inFlight` term — the undatable record — is now measured**, by SecOps: direction is fail-**open** (an undatable record is never in flight, so the guard never holds a project's turns open on a record it cannot date), and that direction is correct, since fail-closed would refuse every turn in a project holding one truncated `status.json`, permanently. One **documented expected survivor**: deleting the term alone leaves the suite green, because `NaN <= C` is already false — it is a defensive restatement, not a live branch under the current spelling.
4. **Single environment.** Every number in this PR is bash 3.2.57(1)-release on one machine with node v24.19.0. Agreement across QA's, BA's, DBA's, SecOps' and Dev's figures is agreement across **checkouts**, not across environments — one observation, five readers. The hook label capture now has three independent captures; all three are on this machine.
5. **`AC12`'s content half is a named human read.** Whether the replacement sentence actually says what it must has no assertion behind it and is not mechanizable without making the test a copy of the comment. `#63-H1/H2/H3` pin the removal and the location only. **This limit has now cost four defects in four drafts:** round-1's claim, round-2's replacement, round-3's parse-divergence clause and round-4's throw-site enumeration each passed all three cells while measurably false. **A green suite has never once flagged this sentence** — every one was caught by a human reading the clause against a fresh measurement. The final version survives by *asserting less*: it states the outcome property that held across every cell anyone ran and deletes every enumeration of the mechanism behind it. When a claim keeps turning out false at the edges, the fix is to stop enumerating the edges, not to enumerate them more carefully. Treat the content half as a human read, always.
6. **Two artefacts are frozen on purpose and must not be tidied.** The assertion label containing `24h` and the `25h` literal on `STALE_ISO`.
7. **NEW — owed to #74, not fixed here.** Three things this round measured that #74 does not yet record: (a) **two** modules read a different column than the guard, not one — `pipeline-status.mjs`'s `?? mtime` substitution is the same grain mismatch #74 records against `session-start.sh`, in a second module #74 does not count; (b) the **parse-function divergence** (`Date.parse` vs `new Date(...).getTime()`), which nothing in this tree pins — swapping the guard to `new Date(...).getTime()` survives all three suites; (c) `pipeline-status.mjs` **throws** (`rc=1`) on a numeric `updated_at` rather than classifying it. All three belong in #74 §1. Off-contract per `status.schema.json` (which types `updated_at` as a date-time string) and outside #63's lane, so reported rather than fixed.

## Incidental finding, not fixed here

While validating the Phase 3 artifact against `plugins/pipeline/schemas/impl-report.schema.json`, one planted defect **survived**: setting a `requirement_check` to `status: "PARTIAL"` with its `justification` deleted validates clean (0 errors) and passes the pre-Phase-4 gate.

The schema declares an `if`/`then` block requiring `justification` when status is `PARTIAL` or `SKIP`, and `dev.md:195` states the same rule. But `validate-pipeline-artifact.mjs` implements only `type`/`enum`/`required`/`minItems`/`allOf` — `grep -n 'schema\.\(if\|then\|else\)'` returns nothing — and `gate-pre-phase4.mjs` has no `justification` or `PARTIAL` handling either. So the rule is enforced by nobody: a written expectation no code reads is a comment.

Not fixed in this PR — changing a shared validator's constraint support is architectural and outside this spec. **Now tracked on issue #66**, which already owns validator inertness (including the knowledge-store under-reporting DBA flagged: the `living-context` entry enumerates the unenforced keyword set as `maxLength`/`pattern` and never names `if`/`then`, which is the one of them a shipped schema actually uses). A deferral parked only in a merged PR body is one step from being a comment, so it has an issue number. (The `AC2` PARTIAL above does carry its justification; it passes on merit, not because the check is absent.)

Related API trap worth knowing: `validate(value, schema, root, ...)` takes the **root schema object** as its third argument. Passing a label string instead makes `deref` fail to resolve and silently skips every `$ref`-declared constraint while still returning `[]` — measured on this very file, a string root also returns 0 errors. The clean result above is meaningful only because the correct call was used *and* because three planted defects each made it go red, and the gate itself was watched failing on a planted copy (`FAIL … exit 1`) and passing on the real one (`OK … exit 0`).

### DIFF
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -506,10 +506,15 @@ function prerequisiteSatisfied(issueDir, row, events) {
  * agree on, because `??` does not fire on it.
  *
  * They also DATE the field differently -- `Date.parse` here, `new Date(...).getTime()` there --
- * which agree on the ISO strings status.schema.json requires and diverge on anything else.
- * `updated_at: 12345` reads here as the year 12345, so the record is permanently in flight,
- * while pipeline-status.mjs never classifies it at all: it throws on the number before its
- * filter is reached. Nothing in this tree pins either spelling.
+ * and MEASURED, those two agree on every STRING, not only the ISO ones status.schema.json
+ * requires: `new Date(string)` delegates to `Date.parse`, so "March 3, 2020", "2020/03/03" and
+ * "not-a-date" all land identically in both. They part only on values that are NOT strings,
+ * where `Date.parse` coerces to string first and the constructor does not. `updated_at: 12345`
+ * reads here as the year 12345, so the record is permanently in flight, while
+ * pipeline-status.mjs never classifies it at all: its filter is not reached on ANY path. With
+ * two or more issue dirs it throws first in `listActive`'s sort comparator, in both markdown
+ * and `--json`; with one dir it throws in `renderMarkdown`; and the sole non-throwing case, one
+ * dir under `--json`, does not run the filter either. Nothing in this tree pins either spelling.
  *
  * That mtime-for-updated_at substitution is the same grain mismatch #74 records against
  * session-start.sh, in a second module #74 does not yet count. No symbol is shared either, so
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -23,8 +23,9 @@
  * among the issue dirs. Two consequences are load-bearing and neither is hypothetical:
  *   - The Stop hook is PROJECT-scoped, not run-scoped, so without a recency ceiling an
  *     abandoned run parked at a guarded phase would refuse every turn in that project forever.
- *     Hence the in-flight predicate below (R6), reusing pipeline-status.mjs's own 24h /
- *     no-final-verdict definition rather than inventing a second one.
+ *     Hence the in-flight predicate below (R6). Its 24h / no-final-verdict window is a SECOND
+ *     copy of the one pipeline-status.mjs holds at :156, not a shared symbol: see the drift
+ *     note above `inFlight`, which is where that duplication lives and where #74 tracks it.
  *   - An explicit signal (CLAUDE_PIPELINE_ACTIVE_ISSUE / PIPELINE_ACTIVE_ISSUE) must not be
  *     able to NARROW the subject: pointing it at a satisfied dir would be the env-var opt-out
  *     the design rejected, and it would leave no trace in the archived record. So both the
@@ -160,7 +161,7 @@ export const UNGUARDED = [
  */
 export const TERMINAL = ["5-archived"];
 
-const IN_FLIGHT_MS = 24 * 60 * 60 * 1000;
+export const IN_FLIGHT_MS = 24 * 60 * 60 * 1000;
 
 /**
  * An unusable risk_tier resolves to the STRICTEST row, never the loosest. That default is right
@@ -487,10 +488,17 @@ function prerequisiteSatisfied(issueDir, row, events) {
 }
 
 /**
- * A run is IN FLIGHT when it was updated under 24h ago and carries no final verdict -- the
- * predicate pipeline-status.mjs already uses to call a run "possibly stuck", inverted. A record
+ * A run is IN FLIGHT when it was updated under 24h ago and carries no final verdict. A record
  * with no readable `updated_at` is not in flight: the guard cannot date it, and a control that
  * cannot date a record must not hold a project's turns open on it.
+ *
+ * DRIFT RISK, LIVE AND UNRESOLVED, TRACKED IN #74. The window below is this module's OWN
+ * literal. pipeline-status.mjs:156 holds an independent copy of the same number to call a run
+ * "possibly stuck", and the two are inverses on the age term ALONE: an undatable record is
+ * neither stuck there nor in flight here, so neither predicate is the negation of the other.
+ * No symbol is shared, so they agree only for as long as nobody moves one of them, and nothing
+ * in this tree fails if one does. #74 carries the several independent spellings of this window
+ * and the abstention they hide; until it lands the agreement is a coincidence, not a contract.
  */
 function inFlight(status, now) {
   if (status.final_verdict) return false;
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -24,8 +24,9 @@
  *   - The Stop hook is PROJECT-scoped, not run-scoped, so without a recency ceiling an
  *     abandoned run parked at a guarded phase would refuse every turn in that project forever.
  *     Hence the in-flight predicate below (R6). Its 24h / no-final-verdict window is a SECOND
- *     copy of the one pipeline-status.mjs holds at :156, not a shared symbol: see the drift
- *     note above `inFlight`, which is where that duplication lives and where #74 tracks it.
+ *     copy of the one pipeline-status.mjs holds in its `stuck` filter, not a shared symbol: see
+ *     the drift note above `inFlight`, which is where that duplication lives and where #74
+ *     tracks it.
  *   - An explicit signal (CLAUDE_PIPELINE_ACTIVE_ISSUE / PIPELINE_ACTIVE_ISSUE) must not be
  *     able to NARROW the subject: pointing it at a satisfied dir would be the env-var opt-out
  *     the design rejected, and it would leave no trace in the archived record. So both the
@@ -493,12 +494,27 @@ function prerequisiteSatisfied(issueDir, row, events) {
  * cannot date a record must not hold a project's turns open on it.
  *
  * DRIFT RISK, LIVE AND UNRESOLVED, TRACKED IN #74. The window below is this module's OWN
- * literal. pipeline-status.mjs:156 holds an independent copy of the same number to call a run
- * "possibly stuck", and the two are inverses on the age term ALONE: an undatable record is
- * neither stuck there nor in flight here, so neither predicate is the negation of the other.
- * No symbol is shared, so they agree only for as long as nobody moves one of them, and nothing
- * in this tree fails if one does. #74 carries the several independent spellings of this window
- * and the abstention they hide; until it lands the agreement is a coincidence, not a contract.
+ * literal. pipeline-status.mjs holds an independent copy of the same number in its `stuck`
+ * filter, to call a run "possibly stuck", and on the AGE term those two comparisons are
+ * complements. Neither predicate is the negation of the other even so, and the DATABILITY term
+ * is where they part -- not in agreement, as an earlier draft of this comment had it. This
+ * guard dates a record from its OWN `updated_at` and abstains when that will not parse;
+ * pipeline-status.mjs substitutes the status.json FILE MTIME for an ABSENT or NULL one first
+ * (`updated_at: status?.updated_at ?? mtime`). So a record carrying no `updated_at` at all is
+ * NOT in flight here and IS listed "possibly stuck" there as soon as the file is a day old.
+ * Measured, not reasoned. An unparseable STRING is the one undatable spelling the two still
+ * agree on, because `??` does not fire on it.
+ *
+ * They also DATE the field differently -- `Date.parse` here, `new Date(...).getTime()` there --
+ * which agree on the ISO strings status.schema.json requires and diverge on anything else.
+ * `updated_at: 12345` reads here as the year 12345, so the record is permanently in flight,
+ * while pipeline-status.mjs never classifies it at all: it throws on the number before its
+ * filter is reached. Nothing in this tree pins either spelling.
+ *
+ * That mtime-for-updated_at substitution is the same grain mismatch #74 records against
+ * session-start.sh, in a second module #74 does not yet count. No symbol is shared either, so
+ * the numbers agree only for as long as nobody moves one of them, and nothing in this tree
+ * fails if one does.
  */
 function inFlight(status, now) {
   if (status.final_verdict) return false;
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -511,10 +511,11 @@ function prerequisiteSatisfied(issueDir, row, events) {
  * "not-a-date" all land identically in both. They part only on values that are NOT strings,
  * where `Date.parse` coerces to string first and the constructor does not. `updated_at: 12345`
  * reads here as the year 12345, so the record is permanently in flight, while
- * pipeline-status.mjs never classifies it at all: its filter is not reached on ANY path. With
- * two or more issue dirs it throws first in `listActive`'s sort comparator, in both markdown
- * and `--json`; with one dir it throws in `renderMarkdown`; and the sole non-throwing case, one
- * dir under `--json`, does not run the filter either. Nothing in this tree pins either spelling.
+ * pipeline-status.mjs never classifies it at all: its `stuck` filter is not reached on any
+ * path. Markdown mode always throws before it -- in `listActive`'s sort comparator or in
+ * `renderMarkdown`'s table loop, depending on where the offending row lands in readdir order --
+ * and `--json` never calls `renderMarkdown` at all, so the filter does not run there even when
+ * nothing throws. Nothing in this tree pins either spelling.
  *
  * That mtime-for-updated_at substitution is the same grain mismatch #74 records against
  * session-start.sh, in a second module #74 does not yet count. No symbol is shared either, so
--- a/plugins/pipeline/tests/test-gate-phase-entry.sh
+++ b/plugins/pipeline/tests/test-gate-phase-entry.sh
@@ -36,7 +36,184 @@ REPO_ROOT="$(git -C "$PLUGIN_ROOT" rev-parse --show-toplevel 2>/dev/null || prin
 iso_hours_ago() { node -e 'process.stdout.write(new Date(Date.now()-Number(process.argv[1])*3600e3).toISOString())' "$1"; }
 
 FRESH_ISO="$(iso_hours_ago 1)"     # in flight
-STALE_ISO="$(iso_hours_ago 25)"    # past R6's 24h ceiling
+STALE_ISO="$(iso_hours_ago 25)"    # past the in-flight window; #63-P2 asserts that against
+                                   # the value the guard exports, rather than restating it here
+
+# ===========================================================================================
+# #63 -- THE IN-FLIGHT CEILING, READ FROM THE GUARD INSTEAD OF RE-SPELLED HERE (preflight half)
+#
+# THE DEFECT. `IN_FLIGHT_MS` decides whether this guard evaluates a record AT ALL, and until now
+# no cell in this suite pinned either side of its boundary. The measured green window was roughly
+# (1.0h, 25.05h): the constant could fall to 1/23rd of its value with every assertion in the
+# three gate suites still green. TWO FIXED FIXTURE AGES WOULD REPRODUCE THAT DEFECT AT NEW
+# COORDINATES, so both ages below are COMPUTED from the value the module exports, and the only
+# remaining bound on the VALUE is the pair of premise cells at the bottom of this block.
+#
+# THE PATH RULE, STATED ONCE AND NOT RESTATED AT EACH CELL. Three resolutions are in play and
+# mixing them is the collision this section is most likely to be broken by.
+#   - #63-A1/#63-A2 (the export and its value) and #63-D1/#63-D2/#63-D3 (the module-shape pins
+#     the declared survivor rests on) read the CHECKOUT, `$REPO_ROOT/plugins/pipeline/scripts/
+#     gate-phase-entry.mjs`. REPO_ROOT derives from PLUGIN_ROOT, which no SCRIPTS_DIR redirect
+#     moves. Bind those to $GUARD instead and they go RED under the rewritten copy described
+#     below WITH THE CHECKOUT UNTOUCHED -- a red cell about nothing, and one that falsifies the
+#     expected-red declaration below for a reason unrelated to the constant.
+#   - The two off-checkout copies (#63-C1..#63-C3, #63-V1..#63-V3) resolve $SCRIPTS_DIR, which
+#     harness.sh:11 leaves overridable.
+#   - probe43 expands $GUARD at CALL time, so `GUARD="<copy>" probe43 ...` drives a copy for
+#     exactly that call. On /bin/bash 3.2.57 a prefix assignment before a FUNCTION call is scoped
+#     to the call (`G=outer; G=inner f; echo $G` prints `outer`); no subshell, no global rebind.
+#
+# DRIFT IMMUNITY, AND WHAT A REWRITTEN RUN IS EXPECTED TO LOOK LIKE. harness.sh:11 is
+# `SCRIPTS_DIR="${SCRIPTS_DIR:-...}"`, so
+#     SCRIPTS_DIR=<tmpdir copy of the WHOLE scripts dir, constant rewritten> \
+#       /bin/bash plugins/pipeline/tests/test-gate-phase-entry.sh
+# drives a different ceiling with the checkout untouched. At 24h and at 2h THIS suite is fully
+# green -- so the declaration below DISCRIMINATES rather than merely permits. At 48h EXACTLY TWO
+# cells in THIS suite are expected red, and both fail for the same correct reason, that STALE_ISO
+# is a 25h LITERAL and 25h is legitimately INSIDE a 48h window:
+#     (i) the existing label `(a) 25h old, no final verdict -> not-applicable even at a guarded
+#         phase`, and (ii) #63-P2 below.
+# Leave those red or record them. Do NOT rebase STALE_ISO onto the constant (out of scope) and do
+# NOT delete #63-P2. Neither red is evidence against the derivation.
+#
+# SCRIPTS_DIR IS NOT TRANSPARENT, AND THE NON-TRANSPARENCY IS NOT THIS CONSTANT'S DOING. Measured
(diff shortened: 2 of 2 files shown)


nsmedia-io:

|---|---|---|
| BA | APPROVE_WITH_NOTES | none |
| DBA | APPROVE_WITH_NOTES | none (both Phase 2 blockers discharged and re-measured) |
| DevOps | APPROVE | none |
| Dev | APPROVE_WITH_NOTES | none |
| SecOps | REQUEST_CHANGES → discharged | one prose clause + a missing deferral; both closed |
| QA | REQUEST_CHANGES → discharged | the same prose clause |

**Final verdict: APPROVE_WITH_NOTES.** Two Phase 2 rounds, a convergence-budget scope-down, three Phase 4 fix rounds and two delta re-reviews. QA and SecOps were re-dispatched unconditionally on every delta round.

### What shipped

`IN_FLIGHT_MS` is exported, and the boundary is pinned by a pair whose fixture ages are **computed from the constant**, so the test cannot rot when the constant legitimately changes. Before this, the measured green window was roughly (1.0h, 25.05h): the constant could fall to 1/23rd of its value with all 558 assertions green.

Runtime behaviour is byte-identical. SecOps swept **14,560 cells** (phases × tier spellings × ages × artifact and verdict states) against both refs and found **0 differing cells**, with a non-zero control (moving the ceiling to 2h) producing 390. DBA replayed every committed `.pipeline/*/status.json` blob across all history: **0 of 50** decision changes, with a re-anchored control giving 15 and 27.

### The finding worth carrying past this PR

The drift comment corrected here shipped a **measurably false claim four times in four drafts, and the suite was green every time.**

1. Claimed the guard reused `pipeline-status.mjs`'s definition. No shared symbol exists.
2. Claimed an undatable record is neither stuck there nor in flight here. `pipeline-status.mjs` substitutes file mtime for an absent `updated_at`, so both fire on one record.
3. Claimed the two date functions "diverge on anything else." `new Date(string)` delegates to `Date.parse`, so they agree on every string and part only on non-strings.
4. Claimed 2+ dirs always throw in `listActive`'s sort comparator and that one dir under `--json` is the sole non-throwing case. Both false: `localeCompare` coerces its **argument**, so only a bad **receiver** throws, and readdir position governs rather than dir count.

`#63-H1`/`H2`/`H3` pin that the sentence was removed and where the replacement sits, never what it says. Every catch came from a human reading a clause against a fresh measurement.

**What finally worked was asserting less.** The shipped version states the outcome property that held across every cell anyone ran (the `stuck` filter is not reached on any path) and deletes every enumeration of the mechanism behind it. Generalisable: when a claim keeps turning out false at the edges, stop enumerating the edges rather than enumerating them more carefully. A mechanism you cannot pin — here a V8 sort internal — should be described by what it guarantees, not by when it fires.

QA ruled the content half **is** mechanizable, aimed at the proposition rather than the sentence, and specified two checks with their non-zero controls. Recorded on #74.

### Verification

- Suite **34 suites / 2561 passed / 0 failed / 0 skip lines**, no cache. Gate trio 493/0, 37/0, 59/0. Run directly by the orchestrator at the merged SHA, not relayed.
- Both **declared expected survivors** confirmed still surviving and still documented (the guard's `<=` vs `<`, and probe43's `>` vs `>=`), so an all-red battery could not be mistaken for coverage.
- Nothing weakened: label-set `comm -23` shows zero lost at every round.
- CI green on the reviewed head.

### Deferred, and written where it survives

- **#74** — three divergent spellings of the window, the mtime grain mismatch, the `rc=1` crash class on non-string `updated_at`, and QA's two proposed content checks.
- **#66** — `impl-report.schema.json` declares an `if`/`then` requiring a `justification` on a `PARTIAL`; the validator implements no `if`/`then`/`else`, so the rule is enforced by nobody.

Three separate agents disclosed **false greens in their own harnesses** during this review (a mutation that matched zero times leaving an unmutated suite looking clean; an `argv` off-by-one; `$?` clobbered by a command substitution). The orchestrator hit a fourth while verifying the final fix. Each was caught by checking the residue rather than the count.

PR #79, opened 2026-08-22
fix: declare 0-setup in both Lane-1 consumer tables (#53)

### LINKED ISSUE #53: 0-setup is a phase pipeline.md writes that no script knows, and the drift test that would catch it derives a narrower population than the one it checks against
Found independently by **two Phase 4 lenses** (BA and Dev) on #34/#42, and written here per the deferral rule before that PR merges. Out of scope for #34/#42, which is why it is filed rather than fixed there.

## The defect

`plugins/pipeline/commands/pipeline.md` writes **27** distinct `current_phase` literals. The drift test in `plugins/pipeline/tests/test-voice-lint.sh` derives **26** of them, and asserts that every phase it found is accounted for in `scripts/voice-lint.mjs`. That assertion passes. The 27th phase is accounted for nowhere.

Derived mechanically at `d8686bc`:

```
loose  grep -o '"\?current_phase"\?: *"[^"]*"'   -> 27
strict grep -o 'current_phase: *"[^"]*"'         -> 26
comm -23 loose strict                            -> 0-setup
```

`0-setup` is written at `commands/pipeline.md:53`, inside the Phase 0 JSON block, as `"current_phase": "0-setup",`. The strict pattern requires an unquoted key, so it never sees a phase written in JSON form. And:

```
scripts/voice-lint.mjs        : 0 occurrences of 0-setup
scripts/gate-phase-entry.mjs  : 0 occurrences of 0-setup
scripts/dispatch-model.mjs    : 0 occurrences of 0-setup
```

It is in neither `VOICE_MOMENTS` nor `NON_VOICE_PHASES`, and in none of `PREREQUISITES` / `UNGUARDED` / `TERMINAL`.

## Why the existing controls cannot report it

This is the interesting half. `test-voice-lint.sh` has an anti-vacuity control, and it does not help:

```
ok    every phase pipeline.md writes is accounted for in voice-lint.mjs
ok    CONTROL: the phase derivation is non-empty (found 26)
```

The control proves the derivation found *something*. It cannot prove the derivation found *everything*, because both the assertion and its control range over the same narrowed set. A phase the grep cannot see is not an unaccounted phase, it is not a phase at all, so the check is green and honest and blind in the same breath.

`voice-lint.mjs` then returns no failures for a run parked at that phase, which is the documented fail-open path its own docstring warns about at `:32` (an unrecognised phase passes silently rather than loudly).

## Why now

#34/#42 rewrote the phase example list in `schemas/status.schema.json` using the **loose** derivation, so the corrected list contains `0-setup`. That makes the schema description the **only** place in the repository that names this phase, and the schema is prose that nothing validates `status.json` against.

There is a second, sharper irony worth recording so the fix is not made twice. `plugins/pipeline/tests/test-gate-phase-entry-drift.sh:41-44` already ruled against the loose form by name:

> `// The STRICT form, matching tests/test-voice-lint.sh:225-239. A looser pattern picks up`
> `` // `0-setup` from the Phase 0 JSON block, so the two checks would disagree about their own ``
> `// population.`

So the divergence was known, written down, and resolved in the direction that hides a real phase. The comment is correct that the two checks disagree. It draws the wrong conclusion about which one is right.

## What "fixed" looks like

Not settled, and worth deciding rather than assuming:

1. **One shared derivation.** Extract a single `pipeline_phase_literals` helper (`tests/harness.sh`, or the marker-extraction idiom `test-status-schema-contract.sh` already reuses from `test-corpus-union.sh`) and have all three suites consume it. Removes the class, not just the instance. It also forces the question of which pattern is correct, once, in one place.
2. **Decide what `0-setup` is.** Either add it to `voice-lint.mjs`'s tables and `gate-phase-entry.mjs`'s (it is a real phase, briefly occupied at the top of every run), or establish that phases written only inside documentation JSON blocks are deliberately excluded and make the *derivation* say so rather than achieving it by accident of a quantifier.
3. **Re-anchor `test-gate-phase-entry-drift.sh:41-44`** whichever way it goes. Right now it is a live justification for the narrower choice.

Note that (1) without (2) converts a silent blindness into a red build, which is the point, but somebody still has to answer (2).

## Related

- #42, the fourth member of the "hand-maintained phase vocabulary rots" class. This is arguably the fifth, and the first where the *derivation* rather than the vocabulary is what rotted.
- #33, on floors that decay when they are not re-derived. Same family: a control whose population is narrower than its subject.

nsmedia-io:
Closes #53

`0-setup` is the run's first recorded phase and a real occupied one, and it sat in no cell of either Lane-1 consumer table. Both suites that police those tables were green with the defect live, because their phase derivation required an unquoted key (`current_phase: "..."`) while `pipeline.md:53` writes that one assignment in JSON form. A phase the pattern cannot see is not an unaccounted phase, it is not a phase at all: the assertion and its control ranged over the same narrowed set.

## What changed

**`plugins/pipeline/scripts/gate-phase-entry.mjs`** gains a fifth exported partition cell, `PRELUDE = ["0-setup"]`, and one `decideForDir` branch. A record at `0-setup` now reads:

```
.pipeline/999 at `0-setup` is the run's setup step, which precedes every guarded phase.
```

where before it read the fail-open sentence for a phase the table does not know, byte-identical to the one an invented phase gets. The `UNGUARDED` branch is not touched, so all nine UNGUARDED reasons are unchanged by construction rather than by replay (replayed anyway, before and after, with a working non-zero control: all nine byte-identical, and the `0-setup` reason, which this change DOES move, shows the instrument can see a change).

The branch sits **after** the terminal check, and the order is load-bearing: moved before it, a `0-setup` record carrying `completed_at` renders the setup-step clause instead of ` is finished.`, un-finishing every concluded run that ended at this phase. A `PREREQUISITES` row was refused by name: `ENTRY` is derived from `PREREQUISITES` keys and must equal the set of phases pipeline.md names at its `Checkpoint first` sites, and `0-setup` is written at no such site.

**`plugins/pipeline/scripts/voice-lint.mjs`** declares `0-setup` non-voice with its mechanism and its general expiry beside it, exports `VOICE_MOMENTS` and `NON_VOICE_PHASES` so the accounting check can read the TABLE instead of grepping this source (a grep a comment satisfies), and adopts the sibling module's `evalEntry` self-run guard **in the same commit as the exports**. Without that guard, importing the module self-ran its CLI: a marker printed before the import appeared, the one after it never did, rc 0. Every membership assertion built on the new exports would have read a green nothing.

Neither table is frozen, and a one-line comment records why: `Object.freeze(new Set(["a"]))` reports `isFrozen === true` and then accepts `.add("b")`, size 1 to 2.

Runtime behaviour is unchanged at every phase. Both hooks still exit 0 in silence at `0-setup`; what changed is that two tables now DECLARE what they previously did not know, and one of them says something true about its subject.

## The behavioural contract

QA authored and committed the failing contract first (`c6a02f1`, three test files, both modules byte-identical to `origin/main`). Baseline 15 red: drift 66/4, voice-lint 61/10, gate-phase-entry 504/1. All 15 are green here, and no QA assertion was weakened, skipped or deleted.

`AC7` red at HEAD, produced by the shipped anchored strip rather than claimed:

```
FAIL  #53-A4 THE CRITERION: with the record-derived span stripped, the judgement
      clause for `0-setup` is NOT the clause for an invented phase.
      expected: discriminates
      actual:   IDENTICAL CLAUSES (31 bytes each): , which is not a guarded phase.
```

## One cross-lane edit

`plugins/pipeline/tests/test-gate-phase-entry-hook.sh` pins test-voice-lint.sh at exactly 48 assertions. That pin is red before this change and after it, so full-suite green is unreachable without re-baselining it, and the lane that owns the file has merged out. Bumped to 71 per the file's own recorded convention, and the property that convention says the cell actually guards was **verified by diffing the suite** rather than claimed: 26 labels added, 3 removed, net +23. It does not hold in its pure form this time, so the note names all three and where each one's substance went (one renamed, two retired with their instrument, each with a named successor) instead of asserting a pure addition. The note also records that this cell is itself a bare count pin of exactly the class #33 retired elsewhere, and that it would be better as a label-set assertion.

## Checks

`bash plugins/pipeline/tests/run.sh`: 34 suites, 2629 assertions, `failed=0`, exit 0.

Mutation battery, run on the committed fix, each restored from git with `git status --porcelain` verified between runs:

| mutation | outcome |
| --- | --- |
| move the `PRELUDE` branch BEFORE the terminal check | CAUGHT, gate 504/1, sole red is `#53-P1` |
| mistype the exported cell name (`PRELUDEE`, consistently) | CAUGHT, drift 66/4 with `PRELUDE=ABSENT` named; gate stays 505/0, which is why the drift suite reads the NAME |
| remove `0-setup` from `NON_VOICE_PHASES` | CAUGHT, voice-lint 68/3, while the string is still in the source |
| remove the `evalEntry` guard | CAUGHT, voice-lint 62/9, PREMISE red first, sibling positive control still green |
| **reword the PRELUDE clause (expected survivor, QA-declared M16)** | **SURVIVED, as intended.** AC7's binding property is that the clause DIFFERS from the 31-byte fail-open sentence, not that it reads any particular way, and `#53-A3` pins that negative exactly |

## Deferrals, and where closure is tracked

Every deferral this branch makes now names a destination in the code itself, following the convention `gate-phase-entry.mjs` already uses for its duplicated in-flight window ("and where `#74` tracks it"). The full ledger is posted as a comment on #53.

- **HANDOFF C** (one shared Lane-1 derivation helper instead of two copies of the wide pattern) — named in both tri-partition comments, with its binding constraint repeated in place: a shared helper returns the EXTRACTION from pipeline.md and never an asserted set, because the Lane-2 status-schema suite asserts 27 by ADDING `halted-error` under a stated rule of its own.
- **The count-pin re-baseline** in `test-gate-phase-entry-hook.sh` — its note now points at the #53 ledger rather than at "whoever revisits this file".
- **`#80`** (the Phase 0 dirty-worktree decision block being structurally unreachable by voice-lint) — cited at the `0-setup` declaration, because #80 is what makes that declaration's expiry checkable rather than aspirational.

Two measurements from the panel are folded in as comments, both re-run rather than copied:

- **`WIDER` is a floor, not a total.** Four spellings tested, three are misses: a bracket-index assignment, a markdown-bold key and a YAML block scalar are matched by neither pattern and land in EXCLUDED as prose; only the single-quoted-value form with an `=` separator reaches UNCLASSIFIED. What bounds the damage is the bare-word population, since a missed spelling is still counted and still has to be named.
- **The excluded fingerprint is non-injective**, and `pipeline.md` already holds a colliding pair: lines 270 and 284 are both 107 trimmed characters, agree for the first 99, and differ only at `SecOps` vs `DevOps`, past the 72-character cut. Neither carries `current_phase`, so it is latent. The two assertion labels are narrowed to the add/remove form the fingerprint actually supports.

That last commit is **comments only**. `git diff -U0 | grep -vE '^[-+]\s*(#|//)'` over it returns exactly the two `assert_eq` label lines; no expected value, population, count or behaviour moved.

## A note for anyone running the suite locally

`bash plugins/pipeline/tests/run.sh` is **34 suites / 2629 assertions / failed=0** on this branch, and the fresh-checkout case inside `test-issue17-integration.sh` re-runs the whole suite in a clone and reports all green.

In a working tree that also holds this run's own `.pipeline/53/status.json`, `test-status-schema-contract.sh` reads 149/1. That suite is untouched by this branch. Its AC3 corpus-maximum control is anchored to a live record (`the corpus maximum is still the 18-char value this control rests on`, `APPROVE_WITH_NOTES/18`), and this run's checkpoint introduces longer verdicts — `SHARD_NONCONFORMANT` (19) and `SCOPE_DRIFT_DECLARED` (20). Isolated to a single variable: with the comment commit in place and only that one record moved aside, the suite is 150/0. The control needs re-anchoring, not deleting, per its own note; it is out of scope here and is routed to the orchestrator.

### DIFF
--- a/plugins/pipeline/scripts/gate-phase-entry.mjs
+++ b/plugins/pipeline/scripts/gate-phase-entry.mjs
@@ -162,6 +162,29 @@ export const UNGUARDED = [
  */
 export const TERMINAL = ["5-archived"];
 
+/**
+ * The run's setup step, and the one phase that precedes every guarded one.
+ *
+ * A FIFTH CELL rather than a PREREQUISITES row or a widening of UNGUARDED, and both refusals are
+ * load-bearing. ENTRY is DERIVED from PREREQUISITES keys and must equal the set of phases
+ * pipeline.md names at its `Checkpoint first` sites; `0-setup` is written at no such site, so a
+ * row there would break a correct invariant. And UNGUARDED's own reason reads "a halt or rework
+ * state, which is never guarded", which is not true of the step that starts the run: shipping a
+ * reason that is false of its subject is the defect this cell exists to correct.
+ *
+ * MECHANISM. pipeline.md writes `current_phase: "0-setup"` at Phase 0 step 5, before any phase
+ * artifact exists and before the first dispatch, so there is no prerequisite a record here could
+ * satisfy or fail. Phase 0's only full-voice owner-facing decision block is the dirty-worktree
+ * halt at step 1, which runs BEFORE that write, so a turn cannot end parked at this phase in
+ * that halt either -- the same mechanism voice-lint.mjs's NON_VOICE_PHASES declaration rests on,
+ * stated here so the two declarations cannot drift apart in their justification.
+ *
+ * EXPIRY, general rather than the single reversal that suggests it: this cell is wrong the moment
+ * ANY owner-facing decision block comes to sit between the step that writes `0-setup` and the
+ * next `Checkpoint first` write, because a run can then END here rather than only pass through.
+ */
+export const PRELUDE = ["0-setup"];
+
 export const IN_FLIGHT_MS = 24 * 60 * 60 * 1000;
 
 /**
@@ -555,6 +578,12 @@ function decideForDir(issueDir, now) {
   const at = `.pipeline/${name} at \`${phase}\``;
 
   if (isTerminal(phase, status)) return decided("not-applicable", `${at} is finished.`);
+  // AFTER the terminal check, and the order is the assertion's subject rather than a style
+  // choice: measured, this branch placed BEFORE it renders a `0-setup` record carrying
+  // `completed_at` as a setup step, un-finishing every concluded run that ended here.
+  if (PRELUDE.includes(phase)) {
+    return decided("not-applicable", `${at} is the run's setup step, which precedes every guarded phase.`);
+  }
   if (UNGUARDED.includes(phase)) {
     return decided("not-applicable", `${at} is a halt or rework state, which is never guarded.`);
   }
--- a/plugins/pipeline/scripts/voice-lint.mjs
+++ b/plugins/pipeline/scripts/voice-lint.mjs
@@ -98,6 +98,13 @@ export const NON_VOICE_PHASES = new Set([
   // between the step that writes `0-setup` and the next `Checkpoint first` write. Re-derive
   // with `grep -n 'full voice\|decision block\|Checkpoint first\|current_phase' on
   // commands/pipeline.md and read what falls between the two.
+  //
+  // #80 IS WHAT MAKES THAT EXPIRY CHECKABLE RATHER THAN ASPIRATIONAL. The mechanism above says
+  // the halt at Phase 0 step 1 is structurally unreachable by this lint, because it runs before
+  // a phase is ever recorded -- so the one owner-facing moment in Phase 0 is not covered by
+  // anything here, and no failing assertion marks that. #80 tracks it. If #80 changes where
+  // that decision block sits, or gives it a phase of its own, this declaration is the line that
+  // has to move.
   "0-setup",
   "0.5-map", "0.5-map-complete",
   "1-ba", "1-ba-complete",
--- a/plugins/pipeline/scripts/voice-lint.mjs
+++ b/plugins/pipeline/scripts/voice-lint.mjs
@@ -67,7 +67,17 @@ const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
 // defect this plugin keeps re-learning. tests/test-voice-lint.sh now parses every
 // `current_phase: "..."` out of pipeline.md and fails when one is neither listed here nor
 // explicitly declared non-voice, so the table cannot drift from the orchestrator again.
-const VOICE_MOMENTS = {
+// EXPORTED so tests/test-voice-lint.sh can assert SET MEMBERSHIP over the table itself. The
+// check it replaces was `grep -q "\"$phase\"" voice-lint.mjs`, a substring grep over this
+// source that a phase named in a COMMENT satisfies and that cannot tell a table key from a
+// mention.
+//
+// NEITHER TABLE IS FROZEN, and that is a ruling rather than an omission. Measured:
+// `Object.freeze(new Set(["a"]))` reports `Object.isFrozen === true` and then accepts
+// `.add("b")` with size going 1 -> 2, because a Set's members are not own properties. Freezing
+// the Set would report a protection it does not provide, and freezing only the object half
+// would leave a reader assuming both were covered.
+export const VOICE_MOMENTS = {
   "1-ba-open-questions": { decision: true, label: "a blocking open question" },
   "1-ba-rework-required": { scales: true, label: "a veto rework halt" },
   "2.5-design-owner-decision": { decision: true, label: "the design-lock" },
@@ -79,7 +89,16 @@ const VOICE_MOMENTS = {
 
 // Phases that are deliberately NOT voice moments: internal checkpoints the owner never sees.
 // Listed explicitly so the drift test can tell "decided this is silent" from "forgot about it".
-const NON_VOICE_PHASES = new Set([
+export const NON_VOICE_PHASES = new Set([
+  // The run's setup step. MECHANISM: pipeline.md's Phase 0 holds exactly one full-voice
+  // owner-facing decision block, the dirty-worktree halt at step 1, and it runs BEFORE step 5
+  // writes `current_phase: "0-setup"`, so a turn cannot end at this phase in that halt.
+  // EXPIRY, stated generally rather than as the single reversal that suggests it: this
+  // declaration is wrong the moment ANY full-voice owner-facing decision block comes to sit
+  // between the step that writes `0-setup` and the next `Checkpoint first` write. Re-derive
+  // with `grep -n 'full voice\|decision block\|Checkpoint first\|current_phase' on
+  // commands/pipeline.md and read what falls between the two.
+  "0-setup",
   "0.5-map", "0.5-map-complete",
   "1-ba", "1-ba-complete",
   "2-constraints", "2-constraints-complete",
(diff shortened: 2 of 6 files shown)


nsmedia-io:

|---|---|---|
| BA | REQUEST_CHANGES → discharged | deferral ledger not durable; closed by filing #80 and posting the ledger on #53 |
| DBA | APPROVE_WITH_NOTES | none (both Phase 2 blockers discharged and re-measured under mutation) |
| DevOps | APPROVE_WITH_NOTES | none |
| SecOps | APPROVE_WITH_NOTES | none; 0 critical / 0 high / 0 medium |
| Dev | APPROVE_WITH_NOTES | none |
| QA | APPROVE_WITH_NOTES | none |

**Final verdict: APPROVE_WITH_NOTES.** Two Phase 2 rounds, a convergence-budget scope-down, a single design pass, and a comment-only pointer commit.

### What shipped

`0-setup` is now a declared phase in both readers. The guard gains an exported `PRELUDE` cell and one branch **after** the terminal check; `voice-lint.mjs` declares the phase, exports both tables, and adopts the self-run guard its sibling already had. Both Lane-1 derivations are widened to the writer's own quoting forms and pinned to a **26-label set** rather than a count.

Runtime behaviour is unchanged. SecOps re-ran a 25-cell matrix at both refs: 21 of 25 byte-identical, and the 4 that differ are exactly `0-setup`'s non-terminal **reason strings**, with decisions and exit codes unchanged. DBA replayed 70 committed records through both builds: 0 differ, with a control that reports 2 when pointed at a different base.

### The finding, and why the old control could not see it

The previous check asserted that every phase it derived was accounted for, and printed `found 26` as its anti-vacuity control. Both the assertion and the control ranged over the **same narrowed set**, so a phase the grep could not see was not an unaccounted phase, it was not a phase at all. Green, honest, and blind in the same breath.

The replacement classifies **every** `current_phase` occurrence (36) as CLAIMED / EXCLUDED-BY-RULE / UNCLASSIFIED, requires UNCLASSIFIED empty, and asserts the excluded bucket as a **membership multiset** rather than a size, because a size pin is the same decaying-count shape one bucket over.

Two things had to be fixed before that landed, and both were the defect reappearing inside its own fix: the first replacement control found 1 of 7 assignment forms while printing `UNCLAIMED = 0`; and a criterion for the only user-visible change **passed at HEAD with the defect live**, because its own named mutation reproduced the HEAD state exactly.

### Disclosed residuals, measured rather than assumed

- The `WIDER` pattern has its own misses: three of four tested spellings land in EXCLUDED. That is a floor, not a total, and it is noted beside the classifier.
- The two tri-partition copies are uncompared: widening one leaves both suites green.
- The excluded fingerprint is non-injective, and two lines already collide on it (neither carries `current_phase`, so it is latent).

All three are in the ledger on #53 with their measurements.

### The cross-lane edit

`test-gate-phase-entry-hook.sh` pinned another suite's whole population at an exact count. It was red before this change and would stay red after, making a green suite unreachable, and the owning lane has finished and merged out. That file's own convention sanctions re-baselining **with a recorded reason**, so it was bumped and the reason recorded in its own style.

The honest part: the convention's stated property is "nothing removed or renumbered", and it **does not hold** — the diff is 26 added, 3 removed. Each removal is named with its successor, three reviewers independently re-derived the diff, and SecOps proved the strengthening live by planting a phase present in the module **only as a comment**: the old suite went fully green at 48, the new one reddens.

### Verification

- **34 suites, 2629 passed, 0 failed**, no skips. Changed suites: drift 70/0, voice-lint 71/0, gate 505/0, hook 59/0.
- One declared expected survivor (rewording the new clause), confirmed still surviving and still exactly one.
- Atomicity verified both halves: widen-only leaves `unassigned ["0-setup"]`, declare-only leaves `notInMd ["0-setup"]`.
- CI green on the reviewed head.

### Deferred, and written where it survives

- **#80** — the Phase 0 dirty-worktree decision block is structurally unreachable by voice-lint, which is the mechanism this change's expiry condition rests on.
- **#53's ledger comment** — HANDOFF C with its binding constraint (a shared helper must return the extraction, never the asserted set, or it deletes another suite's `halted-error`), the three residuals above, and the count-pin re-baseline.

Issue #76, opened 2026-08-22
Mechanical seat for the architecturalTriggers.paths tier trigger (the diff-side half of #21)


nsmedia-io:
Split out of #21 at Phase 1 intake because it lives entirely in files a parallel-session ownership contract froze for #21: `commands/pipeline.md`'s `### Mis-tier tripwire` section and all of `plugins/pipeline/tests/`.

## What #21 does and does not cover

#21 fixes the INTAKE half inside its boundary: BA reads `architecturalTriggers` at Phase 1 and treats `paths` as a mandatory architectural trigger over the ASK, the orchestrator's post-BA validation gains the same trigger, and `commands/pipeline.md` stops implying a path predicate evaluates a diff. It also adds an agent-reported backstop in the owned agent contracts.

That backstop is a NORM (per #40's ruling: a rule stated only in prose is a norm). It does not cover the case it exists for: **scope drifting into `pipeline.config.json` in a diff whose ask never named the file.** At the standard tier that diff runs with no Phase 2 fan-out, no Phase 2.5 and no QA-first Phase 3, and nothing mechanical notices.

## Measured, at 7a052d3

`plugins/pipeline/scripts/data-layer-surface.mjs` against `["pipeline.config.json"]`:

```
["pipeline.config.json"]                     tripwire: false dataLayer: false infra: false
["supabase/migrations/001_x.sql"]            tripwire: true  dataLayer: true  infra: false   <- non-zero control
[".github/workflows/tests.yml"]              tripwire: false dataLayer: false infra: true    <- non-zero control
```

And zero runtime readers of any `architecturalTriggers` sub-key:

```
grep -rn 'architecturalTriggers\.\(paths\|domains\|keywords\)\|architecturalTriggers\[' \
  plugins/pipeline/scripts/ plugins/pipeline/hooks/    ->    0 lines
```

## The ask

Add a fail-closed check over the changed-path list already in hand at the Phase 3-to-4 transition, against `architecturalTriggers.paths`, on the same three-outcome shape as the mis-tier tripwire (hit / clean / INDETERMINATE-halts). Config may only ever WIDEN it, never narrow it, for the reason the tripwire's own union states.

## Why it could not be folded into #21

- `### Mis-tier tripwire` is the correct seat and was forbidden to that session.
[...]

PR #86, opened 2026-08-22
Phase 4 panelists: state when isolation is owed, what qualifies, and what it does not enforce (#19)

### LINKED ISSUE #19: Phase 4 panelists share one worktree, so mutation batteries corrupt each other
Found during the #17 Phase 4 panel (2026-08-19).

\`commands/pipeline.md\` puts all six Phase 4 reviewers in ONE worktree. That is fine for read-only review, but several lenses legitimately run leave-one-out mutation batteries, which edit a real tracked file in place and restore it. Four reviewers hit this independently:

- QA planted a mutation, measured three preset rows as dead cells, then found `DEFAULT_MIGRATION_GLOBS.length` went 14 -> 15 with no checkout of its own. Re-run in an isolated worktree, all 15 rows were load-bearing. Its first measurements were wrong in the under-counting direction.
- SecOps captured `data-layer-surface.mjs` losing a different glob row on two separate observations, advancing monotonically (the leave-one-out signature), and rated it HIGH because a `git commit -a` during a fix round would ship a narrowed halting control.
- DBA and Dev both observed the same file transiently dirty and correctly declined to `git checkout --` it, since that would discard another agent's in-flight work.

No defect shipped: the tree was verified clean and the table intact at 15 rows before merge. But the design makes this reachable every panel.

Options: give each panelist an isolated worktree; or require mutation work to happen on a copy in a tmp dir (which this change's own `test-config-doctor-surfaces.sh` already does correctly); or assert `git status --porcelain` is empty before any Phase 4 fix commit and stage explicit paths only.

nsmedia-io:
Closes #19.

## The problem, observed four times in one panel

Phase 4 puts all six reviewers in one worktree. Several legitimately run leave-one-out mutation batteries that edit a tracked file in place and restore it. In a single panel: QA measured three rows as dead cells and a glob count moving under it with no checkout of its own, then re-ran isolated and found every row load-bearing (its first measurements were wrong in the **under-counting** direction); SecOps watched a surface module lose a different glob row across two observations, advancing monotonically, and rated it HIGH because a blanket commit during a fix round would ship a **narrowed halting control**; DBA and Dev both saw a tracked file transiently dirty and correctly declined to restore it, since that would discard another agent's in-flight work.

**It happened again while this issue was being specified.** A reviewer capturing a baseline watched a concurrent process modify the file this change edits, and self-restore about a minute later leaving no trace. It killed its own measurement rather than trust it. Later, five concurrent full-suite runs were observed in one checkout.

## What ships

Four rules, as one byte-identical block appended to each of the nine agent contracts, plus a paragraph in the Phase 4 preamble:

- **The medium.** A tree qualifies as isolated only when `git rev-parse --absolute-git-dir` differs from the dispatch tree's **and** `git ls-files` exits 0 with a non-zero count.
- **The trigger.** Isolation is owed when a reviewer is about to write a tracked file, not for read-only review.
- **Read-only.** The dispatch worktree is read-only for panelists.
- **Commit hygiene.** No blanket commit; stage explicit paths; record a `git status --porcelain` read.

## Two things measurement changed

**The obvious remedy admits the harm.** "Work in a copy" was one of the issue's three suggestions. A `tar` copy that preserves a linked worktree's `.git` **file** measures identically to correct isolation — same git dir, 132 tracked files, suite reads 85/0 — while sharing the dispatch worktree's index, HEAD and branch, so a commit inside it writes to the tree it was meant to protect.

**And the obvious spelling of the check does too.** `git rev-parse --git-dir` prints a relative path in a main checkout and an absolute one in a linked worktree, so it falsely **refuses** a real clone and falsely **admits** a plain subdirectory of the dispatch checkout. The absolute spelling is correct in all ten cells across both topologies. Verified with a 30-cell matrix (3 spellings x 2 topologies x 5 cells), with the decisive case proved by comparing the inode of the index rather than asserting the mechanism: different in a main checkout, identical in a linked worktree.

That second finding also corrected the acceptance criterion, which had demanded a verdict that is only correct in the topology it forbids.

## What this does NOT claim

**Nothing mechanical enforces any of it.** The SubagentStop validator is separately known to be inert under namespaced agent types and an empty plugin root (#66), which is the default deployment. The block says so.

**The convention's measured compliance record is zero.** Eleven leftover isolation trees on disk, 26.2 MiB, none reclaimable, while the archived attestations for those sessions say the trees were removed. One was created five minutes before a reviewer measured it, by another reviewer, during the round that writes this rule. The prose states the residual rather than claiming a control, and no sentence in the diff says the rule *prevents* anything.

## Scope, narrowed by owner ruling

The schema attestation is **deferred to #76** with its full measurement block: the invalidation table over 52 blocks, the 40-block durable corpus, and the disclosure requirements. Reason: `panelVerdict` has no `additionalProperties` key, so an attestation can be written today and validate cleanly with **no schema change at all**, and the merge step passes whole blocks through. Its entire net contribution was a type constraint plus a description, against a field whose silence would read as compliance.

`config-doctor.mjs` was removed from the editable surface after three independent agents established it contains nothing this convention touches.

## Verification

- `plugins/pipeline/tests/run.sh`: 34 suites, per-suite counters **identical** to the pinned baseline. Run in an isolated worktree built by the rule this change ships.
- A 67-cell battery, all passing, with every control firing and one documented expected survivor (the frozen digest must be blind to an edit outside its span).
- Frozen span digest unchanged across all ten files; the terminator anchor still occurs exactly once per file.
- The nine new copies hash to one digest.
- Worktree registry left as found at 18.

## Known exposure, stated rather than glossed

The nine new copies are held together by **nothing in the repository**. There is no digest test over them and won't be until #76; the battery that checks them is gitignored and does not ship. A one-word divergence in one contract is silent. Falsifiable by hand, not enforced.

Four criteria are keyed to exact literals inside prose, each verified at zero occurrences before this change. A copy-edit breaks a criterion with no visible symptom.

---

## Record: the locked design was amended during Phase 4, deliberately and not silently

The Phase 2.5 design artifact carries a canonical copy of the replicated block, and one verification cell compares the nine shipped copies against it. The Phase 4 panel **ordered a text change** (the location clause named a blessed path as a worked passing example, which is true only on hosts where that variable points somewhere private). The implementer therefore amended the canonical block rather than letting the cell fail against a design the panel had overruled.

- canonical block before: `151327c8aa961a47cfd574c95904d8dacd82739a`
- canonical block after: `3d4681d7cf9a2f1b5e35bced3317ca2539200966`

Both derived the **same way**: extract from the block heading to end of file, keep the trailing newline. An earlier revision of this note recorded the before-value as `cce1b7b2933a577959fcbf9c597c08b63eeb0175`, which is the same block with the trailing newline stripped. The two were therefore not comparable, and a reader re-deriving both under one extraction would have concluded the record was wrong when it was only inconsistently normalised. Corrected after DBA raised it in the delta round.
- canonical block after: `3d4681d7cf9a2f1b5e35bced3317ca2539200966`
- the nine shipped copies hash to that same value, one group, re-derived under three different extraction spellings

**This is recorded here because the artifacts that hold it do not ship.** Only `status.json` is tracked under `.pipeline/19/`; the design and implementation reports are gitignored and vanish with the worktree. So a green comparison cell now means *"the shipped copies agree with the canonical block as amended in Phase 4 by the author of the shipped text"* — which is a sentence a hash comparison can never say for itself, and which would otherwise have been lost.

Raised by SecOps in the delta round as the amendment record being less durable than the thing it guards.


nsmedia-io:
### Delta round, `cfffc69` — design amendment #2, and the DBA D3 normalisation correction

Recorded here because `.pipeline/` is gitignored (`.gitignore:23`), so `design.json` and `impl-report.json` do not survive the merge. The PR body's existing amendment record covers round 1 and is unchanged.

**Amendment (DBA D1).** The locked `canonical_agent_contract_block` is amended a second time: in `**Where to put it.**` the `;` joining the `pwd` assignment to the walk becomes `&&`, plus the prose that says why. Without it a failed `cd` left `p` empty, `dirname` cycled on `.` forever and `/` was never reached (`timeout 3` → exit 124; the same one-liner against an existing path → exit 0). `design.json`'s `chosen_approach.canonical_agent_contract_block` and `preamble_paragraph_1_isolation` were updated to the shipped text, so a green `ACblk.2` means "the nine copies agree with the block **as amended in Phase 4**", not "shipped matches what Phase 2.5 locked". Only the amendment record can tell you the difference.

**One extraction for every sha1 below** — raw, **including the trailing newline**:

```
block: sed -n '/^## Phase 4 tracked-write isolation$/,$p' plugins/pipeline/agents/qa.md | shasum | cut -d' ' -f1
p1:    grep -F '**Phase 4 tracked-write isolation.**' plugins/pipeline/commands/pipeline.md | shasum | cut -d' ' -f1
```

| | pre (`eb040ee`) | post (`cfffc69`) |
|---|---|---|
| nine-copy block | `3d4681d7cf9a2f1b5e35bced3317ca2539200966` | `dca27f7affc21fba08763e1eb3c9e605644de004` (9/9, one group) |
| pipeline.md preamble p1 | `34435409817d2b3b0987c27b4979d283e573effa` | `34db7dc2b5c75793d4159a2651ed2d26984e927f` |

Frozen ten-copy span unchanged at `14b65c48479dfceefb780689adccfbd53656b21e`, one group over all ten files; `replicated verbatim` still exactly 1 per file.

**DBA D3, corrected in `design.json` and `impl-report.json` as well as in the body.** Round 1's record paired two normalisations: the published pre-value `cce1b7b2933a577959fcbf9c597c08b63eeb0175` is the same `2cfcf14` block with its trailing newline **stripped**, set against a **raw** post-value. Two ends under two normalisations cannot be diffed — a reader cannot separate a real edit from the newline. Under the single extraction above the round-1 pair is `151327c8aa961a47cfd574c95904d8dacd82739a` → `3d4681d7cf9a2f1b5e35bced3317ca2539200966`. The mis-normalised number is kept and explained rather than dropped. The preamble p1 pair as first written (`1b56d1c3…` → `ee1a308c…`) was *not* mixed — both ends newline-stripped — and reads `c5397f86e102cd34470b91760322425e5cf125a3` → `34435409817d2b3b0987c27b4979d283e573effa` under the one extraction.

Issue #122, opened 2026-09-02
The empty string still satisfies every required free-text field on the review artifacts (#71 property 3, deliberately left open)


nsmedia-io:
Split out of #71 rather than folded into it, so the deferral is written down before the change that defers it merges. #71 closed its **credential-exposure** half (properties 1 and 2): a write-time refusal in `archiveIssue` derived from `ARCHIVE_ARTIFACTS`, plus the no-credential rule stated on all 21 free-text fields across `review.schema.json` and `peer-review.schema.json`. This is #71's **property 3**, which that change deliberately did not close:

> Whatever distinguishes a present-but-empty value from a meaningful one is applied at the moment of writing, by something that runs, and its failure to run is distinguishable in the output from a clean pass.

## The gap

`review.schema.json` requires `remediation` on every `vulnerabilities[]` entry and `must_satisfy` on every `concerns[]` entry, but `""` satisfies both: the walker in `scripts/validate-pipeline-artifact.mjs` implements no `minLength`, and the requirement cannot be conditioned on severity. A *missing key* is genuinely refused; a present-but-empty one is not. Same actor, same outcome, one keystroke cheaper.

By `evidence.md`'s ship-or-block line this is a **gap, not a ratchet** — a live input defeats it.

## Why #71 did not close it, stated so the next reader does not have to re-derive it

The only seat for a write-time non-blank check is `validate-pipeline-artifact.mjs`, and **#66 records that validator as inert under namespaced agent dispatch, which is the shipping default.** A control placed there fires in the bare-`.claude/agents/` deployments and does not fire where most readers run — the "name the event, name the environment where it occurs" failure. Closing property 3 in that seat would produce a control whose green is a fact about the deployment mode rather than about the record. That is a judgement about *siting*, not a judgement that the gap is acceptable, and it should be revisited the day #66 closes.

`minLength` on the schema has the same problem plus a second one: the repo already rejected that instrument once (`groundOpenQuestions` is a bespoke non-blank check whose own comment records that `""` becomes the cheapest valid value), and the shipped walker would silently ignore the keyword, producing a written invariant with no enforcement.

## Measured, so a later reader can tell whether the population moved

Over this repository's committed `knowledge/issue-archive/*.json` at `fbc5212` (15 archive files):

| row type | rows | field missing | field empty-string |
|---|---|---|---|
| `vulnerabilities[]` / `remediation` | 16 | 0 | 0 |
[...]

Issue #52, opened 2026-08-20
status.json's secret-redaction rule covers ask_text only, while four other free-text fields are committed and archived verbatim


nsmedia-io:
Filed by SecOps as the routed deferral from the #34 Phase 4 panel (reviewed commit `d8686bc`). #34 bounded `events[].verdict` and `flags[].verdict` at 32 chars. This is the half that was deliberately NOT bounded, and the reason it needs a different instrument.

## The exposure

`status.json` reaches a public tree twice:

1. It is the one `.pipeline/` artifact committed to git (`commands/pipeline.md`, durable-checkpoint convention: everything else is gitignored).
2. At Phase 5 it is copied **verbatim** into `knowledge/issue-archive/<n>.json`. `scripts/knowledge-store.mjs:11` puts `"status"` in `ARCHIVE_ARTIFACTS`, and `archiveIssue` at `:142-150` does `archive[name] = doc` then `writeFileSync(...JSON.stringify(archive, null, 2))`. No truncation, no redaction, no field filter.

## The gap

`commands/pipeline.md:63` carries the only secret-handling rule for this file, and it is scoped to exactly one field:

> **`ask_text` is a truncated, human-written task summary. It must never carry a secret.** If a /pipeline argument contains a token-shaped substring (API key, Bearer token, OAuth code, password, `.env` line), redact it before writing `ask_text` ... Only the truncated ask, phase events, and 140-char flag summaries are written to `status.json`; no code path copies provider tokens, Bearer tokens, OAuth codes, or database rows into it.

The redaction imperative names `ask_text` twice and nothing else. Four other free-text fields in the same committed-and-archived file are uncovered:

| field | bound | written by |
|---|---|---|
| `events[].note` | none | orchestrator prose |
| `flags[].summary` | 140 (restated to the writer at pipeline.md:92) | orchestrator prose |
| `veto_reason` | none | orchestrator prose |
| `error` | none | orchestrator prose |

The trailing sentence ("no code path copies provider tokens ... into it") is true and beside the point: these fields are written by the orchestrator, which is not a code path. `error` is the sharpest case, because the natural content of an `error` field is *copied machine output* — a failed `gh`/`curl` echoing a URL with a token, a DB connection error carrying a DSN, a stack trace.

Live example of the unbounded shape, in #34's own PR: `.pipeline/34/status.json` `events[0].note` is **600 characters** of multi-sentence prose. Measured at `d8686bc`:

```
[...]

Issue #40, opened 2026-08-19
Reviewers state properties, not implementations: only QA and Dev propose how


nsmedia-io:
Owner decision, 2026-08-19. Behaviour change across every agent contract.

## The rule

**Every agent may flag an issue, identify a problem, or name a shortcoming. Only QA and Dev may propose HOW to solve it, through the TDD contract.**

The line falls between a **property** and an **implementation**:

- **Allowed, and already required:** the defect, its mechanism, the evidence, and **what a correct fix must satisfy** — stated as a property or invariant.
- **Not allowed:** prescribing code, naming a specific implementation, or writing a patch.

A reviewer's property *is* a test contract. That is the handoff: reviewer says "must satisfy X" → QA writes the failing test for X → Dev chooses the mechanism and implements to green. It routes solution design through the two roles the TDD contract already assigns it to, instead of through whichever lens happened to notice the problem.

## Why the line is there and not somewhere simpler

Both sides of it were observed in one run (#30):

**A property, and it was load-bearing.** SecOps: *"the tripwire's glob set must be a UNION with the built-in defaults, so config can only ever widen the halt."* Taken verbatim, correct, and it closed the blocker. DBA: *"strip only what every target dialect agrees is a comment; classify executable on divergence."* That property closed four independent fail-opens. Neither names an implementation.

**An implementation, and this is the thing to stop.** DBA shipped a five-line replacement for the `--` scanner, validated on an isolated copy. It was correct — and it is a data-safety lens designing the fix, which anchors Dev on one lens' view of a module that serves several.

A strict "defect only" rule would have discarded the first two, which were the most valuable findings in that run. Hence the property/implementation split rather than a blanket ban.

## Collision to handle, not ignore

The shared Phase 4 preamble already requires: *"before you demand a guardrail, name the CORRECT work it refuses."* That obliges a reviewer to reason about the fix's consequences. The new rule must not contradict it. Naming what a guardrail would refuse is reasoning about a **property's cost**, and stays allowed.

Likewise `evidence.md`'s ship-or-block line (a control a live input can defeat is a gap; one only a future edit can defeat is a ratchet) is property-level reasoning and stays.

## Surface

[...]

PR #124, opened 2026-09-02
Refuse credential material at the archive write, and state the rule on every archived free-text field (#71)

### LINKED ISSUE #71: Free-text fields on review artifacts archive verbatim with no credential prohibition (the class #52 named but did not close)
Raised by SecOps and corroborated by Dev during the #40 Phase 4 panel, 2026-08-21. Filed rather than folded because the fix lives in `scripts/knowledge-store.mjs`, which a concurrent session owns.

## The gap

#52 closed COMPLETED against `status.json` only. `commands/pipeline.md` scopes the sole secret rule in the repo to that one file. But `review.json` and `peer-review.json` are both in `ARCHIVE_ARTIFACTS`, and `knowledge/issue-archive/` is a tracked, committed directory.

Every one of these free-text fields on a review artifact is copied verbatim into the committed archive with **no credential prohibition and no redaction**:

- `concerns[].description`
- `concerns[].location`
- `notes`
- `compliance_flags[].concern`
- `advisory_notes`

`#40` added an exposure note to the three fields it introduced or re-contracted (`must_satisfy` on both schemas, `remediation` on `review.schema.json`). The fields above got nothing, because they predate it.

## Why this is worth its own issue rather than a line in #38

1. **The class is wider than #52's fix and is now tracked nowhere.** #52 is closed. Nobody works a JSON description, and documenting a deferral inside the artifact being shipped is not filing it.
2. **#40 is itself the evidence the class is live.** A reviewer had to write the same warning three times to discharge three fields. The fourth through Nth fields on the same objects are the same boundary. The asymmetry already drifted between two siblings written in one commit: `must_satisfy` carries the archive-redaction caveat, its sibling `remediation` does not.
3. **`location` and any error-bearing `description` are the sharp cases, for exactly the reason #52 named**: their natural content is *copied machine output*. A pasted DSN or a token-bearing URL lands verbatim in a public tree, and a fix-forward commit does not remove it from history.

## What the redactor actually does today (measured at `origin/main`)

`redactAbsolutePaths` is the only transform on this path. It keys on path SHAPE via `LEADING_SPAN` (`scripts/knowledge-store.mjs:147`), replacing a leading path token only. There is no length check, no pattern check, and no credential handling of any kind. A pasted secret validates clean and is archived as written.

## A second, related residual from the same panel

`review.schema.json` now requires `remediation` on every `vulnerabilities[]` entry, but the empty string satisfies it: the schema walker implements no `minLength`, and the requirement cannot be conditioned on severity. Measured over the deployments where the gate fires: 229 vulnerability rows, 17 with no remediation (info 9, low 4, medium 4; **0 of 3 critical, 0 of 26 high**), 0 currently empty-string.

By `evidence.md`'s ship-or-block line this is a **gap, not a ratchet**: a live input defeats it, and the same actor reaches the same outcome by writing `""`. It ships because a *missing key* is genuinely refused and accidental omission is the realistic failure mode. But the justification shipped in the schema ("the walker implements no minLength") is the reason this repo already rejected once: `groundOpenQuestions` is a bespoke non-blank check whose own comment records that `""` becomes the cheapest valid value on the field the feature rests on. The repo's answer to this constraint was never `minLength`.

The correct work such a control must not refuse, named: an `info` row whose honest remediation is "none required".

## Properties a correct fix must satisfy

1. Every field that is archived verbatim into a committed tree carries the same prohibition, and the set is derived from `ARCHIVE_ARTIFACTS` rather than hand-listed, so a field added later is visibly missing rather than silently absent.
2. A reader of any one such field can tell, from that field alone, whether anything mechanical protects it. Today three fields say "this is a norm, not a control" and the rest say nothing, which reads as protection by omission.
3. Whatever distinguishes a present-but-empty value from a meaningful one is applied at the moment of writing, by something that runs, and its failure to run is distinguishable in the output from a clean pass.

No mechanism is prescribed here deliberately: per the rule #40 introduces, only QA and Dev propose how.

## Not a duplicate

- #52: closed, scoped to `status.json`.
- #59: closed, fixed the redactor's over-broad path replacement (`e7c1bd2`).
- #38: typed provenance and evidence admissibility on `peer-review.json`. Different subject.
- #66: the validator is inert under namespaced agent types. Different mechanism.

nsmedia-io:
Closes #71.

Dispatched to SecOps rather than Dev because the judgement about what counts as adequate at a credential-exposure boundary is a SecOps call. Both halves are closed; the empty-string half of #71 is deliberately **not**, and is filed as #122 with its reasoning before this merges.

## What I measured before building anything

The issue's premise was half right, and the half that was wrong changed the shape of the fix. Both measurements below were taken at `fbc5212`.

**1. The archived copies were already scanned.** AC-52c in `test-status-schema-contract.sh` walks all of `knowledge/issue-archive/*.json` in full, keys and values. Instrumenting that population showed it already reaches every field #71 names:

| field | strings reached in the existing scan |
|---|---|
| `concerns[].description` | 161 |
| `concerns[].location` | 126 |
| `notes` | 62 |
| `compliance_flags[].concern` | 3 |
| `vulnerabilities[].remediation` | 16 |

25,973 strings over 21 files. So "extend the detection pass to the review artifacts" would have been building a second copy of a control that already exists. `.pipeline/*/review.json` is gitignored and never committed, so the archive is the only path to the tree, and that path was covered — **after the fact**.

**2. Nothing ran at the moment of writing.** A `review.json` with four planted credentials, archived through the shipped CLI:

```
$ node knowledge-store.mjs --archive-issue 999 --from ... --root ...
Archived issue #999 -> .../knowledge/issue-archive/999.json
  artifacts: review
  absolute paths redacted: 1
$ grep -o 'postgres://[^"]*\|Bearer [A-Za-z0-9]*\|AKIA[0-9A-Z]*\|DATABASE_PASSWORD=[^"]*' 999.json
postgres://svc:hunter2hunter2@db.internal:5432/app failed
DATABASE_PASSWORD=s3cr3tvalue
Bearer abcdefghijklmnopqrstuvwxyz012345
AKIAIOSFODNN7EXAMPLE
```

Exit 0, all four verbatim. Note the third line of that output: the run **counted a redaction while shipping the secret**. The `location` was `/etc/app.env:12 DATABASE_PASSWORD=s3cr3tvalue`; `LEADING_SPAN` rewrote the leading path and left the `.env` line standing in the same string. A redaction counting a hit is not a redaction covering it, and that is the sharpest case #71 names.

## The mechanism, and why this one

**A write-time refusal in `archiveIssue`, plus the schema statement. Both halves, and they close different things.**

The prevention is sited at the archive write rather than on the writers, for the argument `redactAbsolutePaths` already makes twenty lines above it in the same file: *a rule stated on the writers has to be restated for every future artifact, and a rule stated here does not*. `review.json`'s free-text fields are written by five different subagents from five different contracts, and `advisory_notes` and `knowledge_drift_claims[].evidence` are **archived while being declared in no schema at all** — so no per-field annotation could ever have covered them. That is not hypothetical; it is why the per-field notes are the statement and not the control.

The walk is over the document `archiveIssue` assembles **from `ARCHIVE_ARTIFACTS`**, so #71's property 1 holds by construction: every field of every archived artifact is in scope, including fields nobody has written yet. The test drives that off the list read out of the source, so a name added to `ARCHIVE_ARTIFACTS` tomorrow is covered without the suite being edited.

It scans the **redacted** document, not the raw one, because redaction sits between the artifact and the file and the redacted copy is the bytes that land. That choice is load-bearing and pinned by its own cell (see the battery below).

I did **not** extend AC-52c's population, and did not touch its scanner. It re-derives its own class table independently, which is what makes it an oracle rather than a restatement; sharing an import would have both go quiet together. What is asserted instead is the **seam**: the class names in the shipped table and in AC-52c's must be the same set, so a narrowing of either goes red.

No length cap anywhere. #52 ruled the instrument is CONTENT and not length, and that ruling is re-asserted here as a paired half: every completeness assertion about the field notes has a twin asserting no `maxLength` appeared, because the cheapest way to make a content problem look solved is to cap the field.

## What stays open, said plainly

- **The empty-string half of #71 (its property 3) is not closed.** `""` still satisfies `remediation` and `must_satisfy`. The only write-time seat for a non-blank check is `validate-pipeline-artifact.mjs`, which **#66 records as inert under namespaced agent dispatch — the shipping default**. A control placed there would be green about the deployment mode rather than about the record, which is the "name the event, name the environment" failure. Filed as **#122**, with the measurement (16 vulnerability rows / 0 empty, 230 concern rows / 0 empty over this repo's archive) and the correct work it must not refuse (an `info` row whose honest remediation is "none required"). The stale `tracked in #71` pointers in both schemas are re-anchored to it rather than left dangling.
- **A secret spelled as prose still reaches the tree.** The guard matches enumerated shapes. Stated in the refusal message, in every field note, in the writer's copy, and in the README, so nobody reads the green as more than it is.
- **`knowledge/issue-archive/*.md` and `*.sh` sidecars are scanned by nothing.** AC-52c globs `*.json` and `archiveIssue` writes only `<n>.json`, so the three `*-qa-battery-record.md` and three `*-verify-*.sh` files in that tracked directory are outside both. Measured clean today against all six credential classes with a literal grep; reported rather than folded in, since it is a different population from #71's subject.

## Evidence

**Non-zero control, one cell per class.** Eleven planted strings, each in its own `concerns[]` row so the reported json path names which row fired: all eleven refused. Four negative controls in the same document — a bare git SHA, a SHA in prose, a slash-separated prose run, an ordinary sentence — none reported, and each negative's own row index is asserted absent rather than a bare substring check that would pass vacuously.

**The five fields #71 names, reached by json path**, plus the two that prove the walk is over the document:

```
.review.secops.concerns[0].description [db_url_creds]
.review.secops.concerns[0].location [env_line]          <- the measured case #71 was filed on
.review.secops.notes [bearer]
.review.secops.compliance_flags[0].concern [aws_akid]
.review.secops.vulnerabilities[0].description [github_pat]
.review.design_review.advisory_notes[0] [slack_token]   <- declared in NO schema
.<key> [aws_akid]                                       <- a credential spelled as an object KEY
```

Refused, exit 1, **and no file written — the archive directory is not even created.** Prevention, not announcement.

**Negative control at the artifact grain.** A clean review artifact archives at exit 0, nothing on stderr, and stdout reports `credential-shaped strings: 0 (of 21 strings scanned)`. The denominator is asserted non-zero *and* asserted to move with the document, so "0 hits" can never be a scan that never ran. A control on the derived-population loop too: a non-`ARCHIVE_ARTIFACTS` file produces the *no-artifacts* error rather than a refusal, proving the seven refusals above were the scan and not a guard that refuses whatever it is pointed at.

**The mutation battery, 8 mutations, 7 red, 1 expected survivor declared.**

| mutation | result |
|---|---|
| drop the refusal, keep the scan (fail-open) | 9 red |
| widen `high_entropy`'s mixed-case lookaheads | 9 red |
| drop the `pg://` alias from `db_url_creds` | 9 red |
| trim a class from the shipped table | 5 red (via the seam) |
| scan the RAW archive instead of the redacted bytes | 2 red |
| drop the object-KEY walk | 2 red |
| strip the note off one schema field | 2 red |
| revert the writer-copy widening | 1 red |
| mutate the `p \|\| "<root>"` path fallback | **SURVIVES — declared** |

The survivor is a theorem, not lost coverage: `archiveIssue` always builds `archive` as an object, so a string leaf is never at path `""` and that branch is unreachable from the only caller that writes a file. Verified directly — `findCredentialMaterial("AKIA...")` returns `<root>`, `findCredentialMaterial({a:"AKIA..."})` returns `.a`. It is recorded in the suite header with the condition that would make it stop being a theorem.

**The RAW-vs-redacted mutation initially SURVIVED**, and that is the most useful thing the battery produced. Fixing it required constructing the case that actually discriminates the two sitings: a token that exists only inside an absolute path redaction replaces wholesale. `/opt/ci/AKIAIOSFODNN7EXAMPLE/build.log` archives clean under the shipped code — correctly, since the written bytes contain no token, and refusing would be a false refusal with nothing in the output to act on — and reddens under the mutation. That cell asserts its own premise (the fixture really did carry the token pre-redaction) so it cannot pass for the wrong reason.

**Over the real committed corpus.** The shipped classifier, run over all 15 `knowledge/issue-archive/*.json` and all 12 committed `.pipeline/*/status.json` — 20 parseable files, 25,581 strings — reports **exactly one hit**:

```
knowledge/issue-archive/34.json .peer-review.secops.concerns[2].description [db_url_creds]
```

Hand-checked, and it is benign: #34's own SecOps shard quoting the `pg://u:p4ssw0rdlong@h/db` it planted as **that review's own non-zero control** while measuring this very exposure. Same single entry AC-52c already allowlists. **No real credential exists in committed history**, so this is a routine fix and not a disclosure.

It is also the named cost of this guard: **re-archiving #34 now throws.** A shape scan cannot tell a fake in a security report from a real DSN, which is why `PIPELINE_ARCHIVE_ALLOW_CREDENTIAL_SHAPES` exists, mirrors `PIPELINE_ARCHIVE_ALLOW_STALE`, and is never silent — it names every hit on stderr, keeps the warning off stdout so `archive-pipeline.mjs` stays a byte-identical re-dispatch, and records the count on stdout.

## Coverage of the statement half

All **21** free-text fields across both schemas now carry the rule — derived structurally (a string-typed leaf with no enum and no `date-time` format), with **no exemption list**, so a field added later is visibly missing rather than silently absent. Each carries what protects it *and* what that protection does not cover, which is #71's property 2.

Three descriptions asserted the opposite and are corrected rather than left: #40 wrote *"there is no length check, no pattern check and no redaction of credential material on this path"* into `must_satisfy` on both schemas and `remediation` on `review.schema.json`. The pattern check now exists. A schema lying about its own controls is worse than one that says nothing, so the correction states which half changed and which did not.

`commands/pipeline.md` — the document the writer actually reads — is widened from `status.json` alone to every artifact reaching the tree, keeping all five of #52's fields named, and distinguishing PREVENTION (nothing is committed) from DETECTION AFTER THE FACT (**amend the commit; do not fix forward**), because they imply different remedies.

## Suite

`bash plugins/pipeline/tests/run.sh`: **all suites passed, 0 failed.** New suite `test-archive-credential-guard.sh` at 103 passed / 0 failed. `test-claims-consumers.sh` AC29 caught the stale `twenty-five` count word in the README's Upgrading lead when item 26 landed — the check working exactly as designed on a hand-copied fact.


nsmedia-io:
## Full suite: 37 suites, all green, `RUNSH_EXIT=0`

`bash plugins/pipeline/tests/run.sh` from `plugins/pipeline/tests/` at `0994566`, against a clean committed tree:

```
All test suites passed.
RUNSH_EXIT=0
```

37 suites run, **0 real failures** (the four `FAIL ` matches in the transcript are assertion *labels* — `AC19: and a failing one as '  FAIL  <label>'` and similar — not results). Key suites: `test-archive-credential-guard.sh` 103/0, `test-status-schema-contract.sh` 154/0, `test-issue17-integration.sh` 135/0 (that is the one whose AC41(c) clones the repo and runs the whole suite *inside the clone*, so the nested tree is green too).

## Two corrections made after the PR body was written

**1. A stale count word in the README, caught by the repo's own check.** Adding Upgrading item 26 left the lead paragraph saying "Twenty-five changes… twenty-five things to know". `test-claims-consumers.sh` AC29 derives the count from the bullets and reddened with `expected: 26 / actual: 25`. Fixed. This is the check working exactly as designed on a hand-copied fact.

**2. The new suite tripped AC10's needle, twice, and the fix went on my side both times.** AC10 in `test-status-schema-contract.sh` walks every *other* suite for a schema file name followed on the same line by one of four present-tense verbs, to catch a comment still asserting that the status schema blesses a phase nothing writes. Its needle cannot tell that subject from mine.

- First draft: the assertion label `"$f.schema.json no longer claims there is no pattern check"`. Reworded so the filename goes last and the verb is outside AC10's class.
- Second draft: **the comment written to explain the first fix** named the verbs literally and reddened AC10 by itself, one edit after the label it existed to explain had been fixed. A scanner cannot tell a quotation from a claim. Rewritten using the bracketed-spelling device AC10's own source already uses on two of its verbs, for exactly this reason.

I did **not** add an exemption to AC10. Its single exemption is deliberate and its own comment states the policy — *"a new suite is covered the day it lands"* — so widening it to make room for a false positive would blunt the check for every suite added after mine.

**Non-zero control on that fix**, because "AC10 is green now" is also what a walk that stopped examining my file would print: a planted `# PLANTED: <status schema file> names 3-scope-drift-adjudication and pipeline.md writes it nowhere.` line appended to the new suite reddens AC10 with `actual: test-archive-credential-guard.sh:581`, naming the planted line number. Removed afterwards.

That restore is worth recording, since `evidence.md` warns about it and I walked straight into it: `git checkout -- <file>` reverted the planted line **and my own uncommitted label fix with it**. It was recoverable only because I had copied the file aside before planting. The commit that carries the fix was made immediately after, before any further mutation.

## Correction to one figure in the PR body

The body says the corpus scan covers "20 parseable files". Precisely: 15 `knowledge/issue-archive/*.json` + 12 committed `.pipeline/*/status.json` = 27 paths, of which the archive-only walk in the reported measurement read 20 (the 12 status records contributed 0 hits and were counted separately in the same run). The finding is unchanged and is the number that matters: **25,581 strings, exactly one hit, hand-checked benign.**

Issue #104, opened 2026-08-31
Two suites pick the resolver under test by sort order off a filename denylist, so the next early-sorting script silently retargets them


nsmedia-io:
Found while adding `dispatch-effort.mjs` in #103; the immediate breakage is fixed there, this is the residual that fix leaves behind.

## The mechanism

`tests/test-dispatch-model-resolver.sh` and `tests/test-mis-tier-tripwire.sh` both discover the script under test like this:

```bash
for cand in $(grep -oE 'scripts/[a-zA-Z0-9_-]+\.mjs' "$PIPELINE_MD" | sort -u); do
  base="${cand#scripts/}"
  case "$base" in
    frontend-surface.mjs|gate-pre-phase4.mjs|...) continue ;;
  esac
  RESOLVER="$PLUGIN_DIR/$cand"
  break
done
```

Discovery-from-`pipeline.md` is deliberate and good: whatever script the orchestrator is actually TOLD to run is the script under test, so a resolver that exists but is documented nowhere reads as unwired. That part should stay.

The problem is the selection rule. It is **`sort -u` order minus a filename denylist, then `break` on the first survivor** — a positional accident, not an identity check. Nothing in either suite states which script it means.

## What it cost

`dispatch-effort.mjs` sorts before `dispatch-model.mjs`. Adding it handed both suites the wrong script. In `test-mis-tier-tripwire.sh` two AC37 rows assert "the resolver EMITS a token" and the effort resolver's agent surface emits nothing BY DESIGN (the Agent tool has no `effort` parameter), so those went red; that then cascaded into `test-issue17-integration.sh`, which runs the whole suite in a fresh checkout and asserts all-green. One root cause, two red suites, and the failure text pointed at the tripwire rather than at the new file.

#103 adds `dispatch-effort.mjs) continue ;;` to both denylists, which fixes the instance and not the class.

## Why the residual is real, not theoretical

The denylist is keyed on names of scripts that exist **today**. Any future `scripts/*.mjs` referenced from `pipeline.md` whose basename sorts before `dispatch-model.mjs` re-enters this trap unannounced. Candidates are not exotic: `archive-*`, `check-*`, `config-*`, `data-*`, `dispatch-*`, anything alphabetically early. The new script's author gets a red suite naming somebody else's file, exactly as happened here.

[...]

PR #113, opened 2026-09-01
feat: enforce the Phase 4 commit-hygiene rule at PreToolUse (#106)

nsmedia-io:
Closes #106

The Phase 4 tracked-write isolation block ended by admitting its own gap: the commit-hygiene rule was "the only one of these rules sited at the actual ship event" and "nothing mechanically enforces it either". Nothing read the prose. This closes it at PreToolUse, the event the rule is actually sited at.

## What ships

**A two-stage gate.** `hooks/pre-tool-use.sh` runs on every Bash tool call and concludes "not a candidate" with shell builtins alone: no forks, no node. Only the narrow plausible-deny branch starts `hooks/pre-tool-use-resolve.mjs`, which answers whose run the call belongs to.

**The forbidden set is a CLASS over flag semantics, not a table of spellings.** The command is tokenized quote-aware, split at top-level operators, and every git invocation in it is judged. `git commit -aqm 'm'` and `git add -Av` stage everything and match no enumerated row, so a literal table would have missed them; a raw substring search would have denied this very PR's doc-retirement commit, whose message quotes `git commit -a`. Both directions are fixture-covered.

**Scoped by ORIGIN, not by role.** The term is the payload's `agent_id`, which Claude Code sets only inside a subagent and omits for the main thread even in `--agent` sessions. So the orchestrator's own status-only checkpoint is exempt by design, and there is no role allowlist — which makes the refusal deliberately wider than the panel, and that over-refusal is disclosed in the shipped text rather than accepted quietly.

**Ownership never comes from mtime.** `resolveRunOwner` narrows to in-flight candidates, then: zero → abstain; exactly one (and datable) → that record; two or more → only an explicit marker naming one of them resolves it, otherwise abstain unconditionally. On a real `git clone --no-hardlinks` of this repository every tracked `status.json` gets a fresh, distinct mtime and the raw resolver returns a record whose own `final_verdict` says its run ended — the case an mtime rule would have got wrong by refusing correct work.

**The predicate moved into a leaf module** so the seam extension does not close a cycle with `gate-phase-entry.mjs`. `IN_FLIGHT_MS` stays exported from the guard in the exact source text `test-gate-phase-entry.sh`'s #63-A2 pins, and is passed into the leaf, so that suite's ceiling-rewrite mutation still bites.

[...]

PR #126, opened 2026-09-02
Refuse an over-cap status.json verdict at write time, from the cap the schema declares (#117)

### LINKED ISSUE #117: status.json's 32-char verdict cap is only checked after the fact by CI, so the orchestrator writes past it repeatedly with no write-time signal
Raised by QA during #106's Phase 4 round 3 (`gh pr view 113`), after the orchestrator's own writes to `status.json` violated the schema's 32-char verdict cap twice in one run.

## What happened, both times, with commits

**First occurrence**, before Phase 4 round 1 even started: 7 verdict labels the orchestrator had written across earlier phases exceeded 32 chars (the longest, `DEV_APPLYING_AC18_AC33_TEST_EDITS_DISPATCHED`, was 44). Found only because `test-status-schema-contract.sh` happened to run and report it; fixed by hand, all 7 shortened.

**Second occurrence, and the one worth fixing structurally**: the orchestrator wrote `"PAUSED_FOR_OWNER_ROUND_3_DECISION"` (33 chars) into `status.json` on its own checkout, then dispatched Dev for a Phase 4 fix round. Dev's own work touched the SAME file in the worktree (this repo's own `.gitignore:24` tracks `.pipeline/*/status.json`, so the file legitimately lives on both the orchestrator's checkout and the feature branch), found the 33-char label as pre-existing red, and fixed it there (commit `6eefeb6`). The orchestrator then did a routine "sync my checkpoint into the worktree" step -- a plain `cp` from its own (still-unfixed) copy over the worktree's (now-fixed) copy -- and committed that as `adce70c`, silently undoing Dev's fix. The reviewed commit for Phase 4 round 3 was `adce70c`. Both QA and SecOps, working independently, caught the same two failing suites and named the same root cause.

## Why this keeps recurring

Nothing checks the cap at the point of writing. `commands/pipeline.md` states the rule in prose ("A TOKEN, not prose: strict 32-char cap... Nothing validates status.json against that schema, so this restatement IS the write-time honorer") and relies entirely on the orchestrator remembering to self-police every edit, for the life of a run that can span dozens of checkpoint commits. The second incident shows the failure mode isn't just "forgot to check a new value" -- it's that a FIX made in one copy of a tracked file can be silently overwritten by a sync of an unrelated, stale copy, and nothing at commit time notices the regression either.

## Scope, left to intake

Not proposing the exact mechanism. Two shapes that occurred to me, for whoever specs this:
- A cheap pre-commit check the orchestrator (or anyone) runs against `status.json` before `git commit`, refusing any `events[]`/`flags[]` `verdict` field over the schema's cap -- this is what QA's own suggestion during round 3 was pointing at.
- Extending `test-status-schema-contract.sh`'s existing walk (it already does exactly this check, just after the fact, in CI) into something offered as a small standalone script the orchestrator can invoke inline before every checkpoint commit, rather than only discovering the violation whenever the full suite next happens to run.

Either way, the goal is the same: catch this before it lands in a commit, not after a Phase 4 panelist spends part of a round rediscovering it.

## Why this isn't blocking #106

Both instances were fixed by hand within #106's own run and independently verified clean. This issue is about making the NEXT run's version of this mistake cheaper to catch, not about #106's own gate behavior.

nsmedia-io:
## What

`status.json`'s 32-char `events[].verdict` / `flags[].verdict` cap was declared in `schemas/status.schema.json`, restated as prose in `commands/pipeline.md`, and enforced by nothing until `test-status-schema-contract.sh` walked the committed corpus in CI -- after the violating record was already in history.

`scripts/check-status-record.mjs` is the write-time honorer. One command, no arguments to remember:

```
node "${CLAUDE_PLUGIN_ROOT}/scripts/check-status-record.mjs"
```

It discovers `.pipeline/<n>/status.json`, reads **each field's own `maxLength` out of the schema** (the digits `32` appear nowhere in its executable lines, asserted), and exits 1 naming the file, the json path, the value and both lengths. Silent, exit 0, when clean.

**It reads FILE CONTENT, not a diff, and takes no argument naming what changed.** That is deliberate and it is #117's second occurrence: a 33-char label was fixed in a worktree (`6eefeb6`) and silently restored when a routine `cp` from a stale checkout overwrote the fixed copy (`adce70c`). A check scoped to "the value I just wrote" passes that. `test-check-status-record.sh` replays the three-step sequence as a regression cell.

**Nothing checked is never a pass.** An empty `.pipeline/`, an absent one, a truncated record, an `events` written as an object, or an explicitly-named file that does not exist all exit **2**, never 0.

## One reader, not two

`test-status-schema-contract.sh`'s inline walker is deleted; its corpus walk now shells out to the same script through `--report`. The CI walk and the write-time refusal are one reader of one cap. The suite keeps everything the script does not own: the tracked/on-disk union population, the vacuity controls, and the fixture matrix.

`--cap` (which those controls use to reach a violation on the live corpus) **may only TIGHTEN**. A value that would loosen either field is refused with exit 2, so the flag cannot silence the check it measures.

## Out of scope, deliberately

No free-text field gains a length opinion. #52 ruled the instrument for `events[].note`, `flags[].summary`, `veto_reason` and `error` is CONTENT, not length. `test-check-status-record.sh` ratchets that: a record whose free-text fields are 2000 characters each must exit 0, with a control confirming its verdicts were still inspected.

## Evidence

**Boundary at exactly the cap.** `maxLength` is inclusive, and the live `#106` record sits at exactly 32 (`SKIPPED_OWNER_DECISION_CONFIRMED`), so `>=` would refuse a record the schema permits. At exactly 32: exit 0, silent. At 33: exit 1. Both pinned.

**Real corpus.** This branch's 11 committed records: `files=11 read=11 verdicts=177 longest=18 (APPROVE_WITH_NOTES) violations=` exit 0. The orchestrator's live checkout, which holds the in-flight `#106` record: `files=12 read=12 verdicts=274 longest=32 (SKIPPED_OWNER_DECISION_CONFIRMED) violations=` exit 0 -- the boundary value passing on the real thing.

**The cap really is read.** Raising `maxLength` in a scratch schema makes the same over-cap record pass; lowering it makes the same at-cap record fail against the lowered number; deleting it exits 2 with `no usable maxLength` rather than passing. All three are suite cells.

**Mutation battery, 8 mutations, 7 red and 1 declared survivor:** `>` to `>=` (8 red), `>` to `> cap+1` (22), absent-verdict counted as a violation (6), cap hardcoded to 32 (4), the two per-field caps conflated (1), empty corpus reads as pass (3), `--cap` tighten-only removed (2). **SURVIVED, expected:** dropping `found.sort()` in discovery -- no assertion depends on discovery order, the report is a set and every violation is enumerated regardless, so the gap changes no verdict.

## Two adjacent repairs this change forced

- `check-status-record.mjs` sorts FIRST among the `scripts/*.mjs` references in `pipeline.md`. That is exactly the "next early-sorting reference" #104's ambiguity halt predicted; it named the script loudly on the first run instead of silently handing the resolver suites the wrong file. Denylisted by name in both discovery blocks, with the reason.
- `test-pipeline-telemetry.sh`'s `worktree_path` assertion was matching `grep -r`'s absolute PATH prefix rather than line content. Identical sources report `0` from a checkout named `agent-pipeline` and `8` from one named `i117-status-check`, the eight hits being in two files that had not changed. `-rhn` scopes the needle to the line. Non-zero control re-taken on the repaired form.

Closes #117


nsmedia-io:
## CI is RED, and the red is inherited from `main` -- this branch adds zero failures

**The identical two suites fail on pristine `main` with this branch absent.**

| | run | result |
|---|---|---|
| `main` @ `4b3577d` (0.35.0, #113's merge) | [33590464767](https://github.com/nsmedia-io/agent-pipeline/actions/runs/33590464767) | `2 suite(s) FAILED.` -- `test-issue17-integration.sh`, `test-pretooluse-doc-retirement.sh` |
| this branch @ `9afdeb3` | [33593681910](https://github.com/nsmedia-io/agent-pipeline/actions/runs/33593681910) | `2 suite(s) FAILED.` -- the same two, same assertions |

**One root cause, four cascades.** The only real failure is `AC29: 'nothing mechanically enforces it either' appears in ZERO files under plugins/pipeline/`, whose actual is `plugins/pipeline/README.md`. #106's own release-note bullet (README.md:117, describing the `PreToolUse` gate) quotes the sentence its own retirement test bans repo-wide under `plugins/pipeline/` with only `tests/` excluded. That is the shape `test-pretooluse-doc-retirement.sh`'s own header names: *"A ban that fires on the diff that retires it is the recorded shape of this defect."* It landed on `main` firing on a README bullet instead.

`test-issue17-integration.sh`'s four failures are downstream of it: AC41(c) clones the tree and runs `run.sh` inside, and the clone reports `RED INSIDE THE CLONE: test-pretooluse-doc-retirement.sh` -- one name, nothing else.

Proof this branch is not involved:

```
$ git show origin/main:plugins/pipeline/README.md | grep -c 'nothing mechanically enforces it either'
1
$ git diff --name-only origin/main...HEAD | grep -c README
0
```

**Not fixed here, deliberately.** It is #106's artifact on a surface a concurrent lane is editing, and rewording another lane's release note to manufacture a green would hide `main`'s red rather than surface it. Whoever owns that lane: either reword the bullet so it does not quote the literal, or give AC29 the quotation-aware exemption the same suite already gives `tests/`.

**Everything this PR touches is green on both platforms**, macOS 26.0 and ubuntu-latest under `PIPELINE_TESTS_REQUIRE_CAPABILITIES=1`, with identical tallies: `test-check-status-record.sh` 70/0, `test-status-schema-contract.sh` 154/0 (the rewired corpus walk), `test-pretooluse-gate-declaration.sh` 67/0 (module count bumped 17 -> 18, which is the pin going red on schedule as its own comment asks it to).

Issue #98, opened 2026-08-31
Settle empirically whether a per-call effort override reaches a named agent


nsmedia-io:
> **RE-SCOPED 2026-08-31.** This body is a render of `.pipeline/98/spec.json`, which is the source. The original mechanism (a `dispatch-effort.mjs` resolver mirroring `dispatch-model.mjs`) is dead: the parameter it was designed to feed does not exist. See q1 below.

> **REMEDIATION ROUND, 2026-08-31 (BA-2, re-review).** This body render is stale against `spec.json` in several specific places (below, patched in place) and was never regenerated after the round-2 fixes. Full current evidence (22 `measured_state` rows, not the 6 below) is in comment 2. The wiring issue this out_of_scope section refers to is now filed for real: nsmedia-io/agent-pipeline#101.

## Problem
RE-SCOPED 2026-08-31, after the original spec's own q1 was settled against it. The original ask was a `dispatch-effort.mjs` resolver mirroring `dispatch-model.mjs`, printing a token the orchestrator would pass as an `effort:` parameter on an `Agent()` dispatch. That parameter does not exist. code.claude.com/docs/en/sub-agents documents a per-invocation `model` parameter with an explicit four-step resolution order and mentions `effort` in exactly four places on the whole page, only one of which is the frontmatter-field table row; no per-invocation effort override is documented for the Agent tool, and the shipped runtime's Agent-tool spawn path reads `effort` off the resolved agent definition (`effort:a.effort` where `a` is `selectedAgent`) with no opts override in the expression. A resolver computing a value with no parameter to receive it is dead code with no consumer, which is exactly the risk q1 named. HOWEVER the ask is not dead: a per-call effort override DOES exist on a DIFFERENT dispatch path. The `Workflow` tool's `agent()` helper accepts both `opts.effort` and `opts.agentType`, and in the shipped implementation the two compose in the direction this ask needs. What is NOT established is that the shipped code path is what actually executes, that a plugin agent of this pipeline resolves as an `agentType`, or that this pipeline's SubagentStop-based artifact validation survives a Workflow dispatch. This issue is now scoped to settling those, and to nothing else. The underlying flat-effort problem is unchanged and real (9 agent contracts, one pinned effort line each, nothing anywhere varies it), and stays open behind this spike.

## Root cause
[...]

PR #129, opened 2026-09-02
fix: make a status.json record mean what it says about being live (#110, #74)

### LINKED ISSUE #74: The in-flight window is spelled independently in five places, one over a different column, and the abstention it governs is invisible in production

### LINKED ISSUE #110: current_phase re-enters a guarded phase carrying a stale final_verdict, so anything gated on final_verdict is blind for the whole fix round
## Found during

#106's Phase 2 round 5 (DBA), while reviewing a rule (`R5`) that excludes a record from in-flight narrowing the instant it carries a `final_verdict`, on the reasoning that `final_verdict` is a positive, self-describing statement that the run is over. Not really about #106's own hook -- pulled out because it's a gap in `status.json`'s own lifecycle, predating this issue and broader than it.

## What's wrong, with a real historical example

`commands/pipeline.md` writes `final_verdict` once, at Phase 4's verdict computation (`:1002` in the current tree). On a `REQUEST_CHANGES` (or `REQUEST_REFACTOR`) verdict, the delta-review flow (`:645`) re-enters `current_phase: "4-review"` for the fix round -- but nothing clears `final_verdict` when it does. So for the entire duration of a fix/delta round, the record sits at a guarded phase (`4-review`) while still carrying the PREVIOUS round's `final_verdict`.

DBA proved this isn't theoretical by scanning this repo's own committed history, not by reasoning about the code in the abstract:

```
c337579  .pipeline/19/status.json
current_phase: "4-review"   final_verdict: "REQUEST_CHANGES"   review_rounds: 2
events tail: [4-review REQUEST_CHANGES "round 1"], [3-impl FIX_ROUND_COMPLETE]
```

That's issue #19 -- the actual issue the Phase 4 tracked-write isolation rule exists to prevent a repeat of -- caught mid-fix-round in exactly this state.

`pipeline.md:1010` compounds it: `current_phase: "4-review-complete"` is written in the SAME update as `final_verdict`, so that specific phase literal can never actually co-occur with an unset `final_verdict` either -- a rule-table entry that looks live but is structurally dead.

## Impact

Any future control (like #106's PreToolUse gate) that reasons "a record with a `final_verdict` is concluded, exclude it" will be silently blind for the entire remediation window of every `REQUEST_CHANGES`/`REQUEST_REFACTOR` issue -- which, per this repo's own telemetry, is where Phase 4 spends the plurality of its active time (29% across seven measured runs). That's not a rare edge case; it's the common shape of a hard issue.

#106 worked around this by keeping its own exclusion narrow and documented rather than waiting on this fix (see its `R5`/`R21` history). This issue is the actual fix underneath that workaround.

## Fix

Clear `final_verdict` (and `peer_review_verdict_counts`, which is derived from it) at the SAME checkpoint write that re-enters `current_phase: "4-review"` for a delta round, so the field's presence genuinely means "this run concluded" rather than "this run concluded at some point in its history." Verify against real committed records the way DBA did here, not just against a fresh fixture -- the multi-round records in `.pipeline/*/status.json` history (or their Phase 5 archives in `knowledge/issue-archive/`) are the actual population this needs to hold over.

## Out of scope

Any change to #106's own PreToolUse gate. #106 ships with a narrower, documented exclusion rule that doesn't depend on this fix landing first.

nsmedia-io:
Closes #110
Closes #74

Two defects, opposite ends of one question: **what does a `status.json` record actually claim about whether a run is live.** Fixing one alone leaves the class open, so they ship together.

## Reproductions, on real committed records, before anything changed

**#110 — a concluded record that is not concluded.** Scanned all 36 distinct `status.json` blobs in all history plus the `knowledge/issue-archive/` copies:

| record | state |
|---|---|
| `.pipeline/19/status.json` blob `d83bfc0` | `current_phase: "4-review"`, `final_verdict: "REQUEST_CHANGES"`, `review_rounds: 2` |
| `knowledge/issue-archive/43.json` (**checked out today**) | `current_phase: "3-impl"`, `final_verdict: "REQUEST_CHANGES"`, `review_rounds: 4` |

The second was found by the new scanner, not by reading code, and **it is what broadened the fix.** That run sat in the stale-verdict state at a guarded phase for a four-round remediation and *never reached a `4-review` re-entry* — so the issue's stated fix (clear on re-entry to `4-review`) would not have caught it. The rule now sits at **both** points, and the earlier one is the loopback to the Dev step.

The cost was already measured in this repo, at `scripts/gate-phase-entry.mjs:697`: *"a refusing record at `3-impl` given a `final_verdict` goes rc 2 -> 0, and every later phase of that run is not-applicable thereafter."* So a stale verdict does not merely blind a control for a window — it disarms the phase-entry guard for the rest of the run. `3-impl` is a guarded phase (`PREREQUISITES`, line 117).

**#74 §2 — mtime vs `updated_at`.** Re-taken in a fresh `git clone --no-hardlinks` pinned at `856a5d0`:

```
.pipeline/106     mtime 0.00h   updated_at   9.99h   4-review          NOTICE-FIRES / guard-IN-FLIGHT
.pipeline/17      mtime 0.00h   updated_at 341.49h   3-impl-complete   NOTICE-FIRES / guard-abstains
.pipeline/39      mtime 0.00h   updated_at 321.40h   3-impl            NOTICE-FIRES / guard-abstains
.pipeline/98      mtime 0.00h   updated_at  37.99h   4-review          NOTICE-FIRES / guard-abstains
exp-airlock       mtime 0.00h   updated_at 325.74h   1-ba              NOTICE-FIRES / guard-abstains
exp-claims        mtime 0.00h   updated_at 327.74h   1-ba              NOTICE-FIRES / guard-abstains

notice fires on 6, guard calls 1 in flight -> 5/6 = 83.3% disagreement
```

Every disagreement is in the **claim-more** direction: the operator is told a run is in flight that the guard already abstained on. The issue measured 4/5 = 80% at an earlier commit; the corpus has since grown by #106 and #98.

**Non-zero control for that harness:** setting each file's mtime to its own `updated_at` and re-running the identical script gives `notice fires on 1, guard 1 in flight, 0 disagreements`. The harness discriminates; the 83.3% is caused by the clone, not by the scanner.

## What was unified, and the proof it bites

`pipeline-status.mjs` held `age > 24 * 60 * 60 * 1000 && !r.final_verdict` — an independent copy, no shared symbol. It now reads a `stuck` verdict derived in `run-candidates.mjs` from the **same `ceilingMs` term** `inFlight` uses. #106's `run-candidates.mjs` was already the leaf both the guard and the PreToolUse gate read; this joins the third consumer to it rather than inventing a fourth arrangement.

The bite is proved, not asserted (`test-pipeline-status.sh`): the suite copies the whole `scripts/` directory, applies a **value-relative** shift to the leaf's `IN_FLIGHT_MS` (computed from the shipped value, never a rewrite of the `24 * 60 * 60 * 1000` literal), prints and asserts the changed line, then requires **both** consumers to flip on the same record:

- **Consumer 1** `pipeline-status.mjs`: `not stuck` -> `Possibly stuck`
- **Consumer 2** `validate-pipeline-artifact.mjs`'s `resolveRunOwner`: `inference` -> `none`

Both have baselines asserted against the shipped module first, so each flip is a flip.

**Mutation expected to SURVIVE, with its reason:** shifting `gate-phase-entry.mjs`'s **own** `IN_FLIGHT_MS` declaration must leave the reporter untouched. That is a theorem about the design, not a coverage gap — the guard passes its constant in explicitly and the leaf's copy is the default for callers with no guard to ask, a separation `run-candidates.mjs` documents and #63-A2 depends on. What keeps the two equal is `test-pretooluse-inflight-ceiling.sh`. **If that cell ever reddens, the separation has been collapsed** and that suite, not this one, is where the drift should have been caught.

## `session-start.sh`: what the new predicate refuses

The notice runs only when the guard is disarmed (node missing, or the script missing), so it cannot use node. It now compares the ISO strings as **integers** (`YYYYMMDDHHMMSS`), tries **BSD `date -v` then GNU `date -d`** in that order — BSD's `-d` means "set DST", so the GNU spelling would misparse on macOS rather than fail — and falls back to the **old mtime test** when neither dialect works or the timestamp is not UTC-Z.

- **It refuses:** announcing a disarmed guard in a checkout where every record is already stale by its own `updated_at`. That is exactly the case where the guard would have abstained on all of them anyway, so the disarm has no live consequence to announce.
- **It newly surfaces:** a record dated in the *future*. The guard has no floor either, so agreeing there is the point.
- **The degraded path is byte-identical to what shipped.** A superset costing noise beats a subset costing the announcement, for a warning about a disarmed safety control.

`updated_at` is UTC-Z in 25 of 25 committed blobs (22 `...SSZ`, 3 `...SS.mmmZ`), which is why truncating both sides to 19 characters is exact. A probe cell **asserts** (not prints) that the runner's `date` produced a cutoff, so a Linux box where the fallback silently took over fails loudly instead of measuring the wrong path.

## Disposition of committed historical records

**Not migrated.** `.pipeline/*/status.json` history is the audit trail of what the orchestrator actually wrote, and it is the evidence that reproduced both defects. The new corpus check reads the **working tree only**: it asserts **zero** over `.pipeline` (the surface the fix controls, and it is zero today) and **pins** the archive's one known record by name rather than excluding it, so a *new* archived record in that state reddens. `knowledge/issue-archive/43.json` is left for the Librarian, who alone may write the knowledge store.

## The structurally dead rule-table row, decided

`gate-phase-entry.mjs`'s `4-review-complete` row cannot fire: `pipeline.md` writes that phase literal in the same update as `final_verdict`, and the guard only evaluates records `inFlight` admits, which requires no verdict. **Measured, not reasoned:** all 7 records ever committed at that phase carry a verdict; 0 do not.

**Kept**, for two reasons — it is the exit half of an entry/exit pair the `GUARDED` derivation reads from this one table, and the unreachability is a property of `pipeline.md`'s write rather than of this file. The deadness is documented at the row, and **its premise is asserted**: if the two writes are ever split, the test reddens and says the row has come *alive* and the comment must be rewritten, rather than asking anyone to put the old behaviour back.

## Stale-claim sync

The `#110` disclosure replicated byte-for-byte across nine `agents/*.md` and `commands/pipeline.md` said *"this hook does not fix it."* Left alone, the repo would ship a doc claiming an open defect with a test enforcing the claim. Updated in all ten: the state is now closed **at its source** while the hook's own exclusion is unchanged, so a record written by an older orchestrator or by hand still lands in the same silence. The nine remain byte-identical (verified: 1 distinct digest), `pipeline.md`'s stays a distinct compressed restatement, and all ten still carry the `#110` token AC30(f)'s clause detector looks for. AC30(f)'s behavioural assertion is unaffected — the gate's exclusion rule did not change.


nsmedia-io:
## CI round-trip: four failures, all mine, none closed by softening a check

First push was green locally and **red on ubuntu** — 7 suites. Controlled comparison first: all 7 pass at `856a5d0` in a pristine clone, so they were mine and not contention. Four distinct causes.

**1. `test-run-lifecycle.sh` — a real portability defect in my own suite.** A premise cell grepped a BRE containing `` \` `` to match a literal backtick. In a basic regular expression `` \` `` is a **GNU extension meaning start-of-buffer**, so the pattern matched nothing under GNU grep on ubuntu while matching a literal backtick under this host's `ugrep`. Same pattern, same file, opposite answers. Now `grep -nF` with single quotes, so neither the shell nor the regex engine gets an opinion.

**2/3. `test-dispatch-model-resolver.sh`, `test-mis-tier-tripwire.sh` — #104's ambiguity halt, working as designed.** Worth stating precisely, because the natural guess is wrong: **no new `scripts/*.mjs` was added.** `run-candidates.mjs` has existed since #106; what is new is its first *reference* from `pipeline.md`, added by #110's clearing rule. Excluded **by name** in both denylists with the reason recorded, exactly as #117 did for `check-status-record.mjs`. The ambiguity check itself is untouched.

**4. `test-pretooluse-doc-retirement.sh` — AC33 caught a bad citation of mine.** `scripts/gate-phase-entry.mjs:697` does not resolve from the repo root. Their control was right and my citation was wrong; now `plugins/pipeline/scripts/gate-phase-entry.mjs:697`, matching the convention the only other citation in that file uses. Line 697 re-verified as the "Measured: a refusing record at `3-impl`" sentence after my own edits shifted that file.

**5. `test-voice-lint.sh`, `test-gate-phase-entry-drift.sh` — the #53 tri-partition pin, RE-DERIVED not softened.** The EXCLUDED bucket is a fingerprint multiset over every bare-word `current_phase` occurrence in `pipeline.md`. #110's two new paragraphs each add one **prose** mention, so the bucket legitimately goes 8 -> 10 occurrences, 5 -> 7 unique. The load-bearing property is unchanged and still asserted: **UNCLASSIFIED is EMPTY**, so no real assignment escapes the derivation. The new value was produced by running the suite's own `tripart` against the tree, not hand-copied from a failure message, and both copies were updated identically including the prose stating the collapse ratio.

**Proof the re-derived pin still bites** (a pin edited until it matches is an assertion following the code): appending one ghost prose line carrying `current_phase` to `pipeline.md` reddens **2 cells in each copy** — `test-voice-lint.sh` 272/0 -> 270/2, `test-gate-phase-entry-drift.sh` 70/0 -> 68/2. Restored from **git**, not from memory, with the tree verified clean afterwards.

`test-gate-phase-entry-hook.sh` needed no change: it pins voice-lint at exactly 272 assertions, and that count is restored.

**One pinned entry now carries an apostrophe** (`panel's`), spliced as `'\''` because the block is a single-quoted shell literal. That is the fixture's limitation being handled rather than the documentation reworded to suit it — noted in place, since future `pipeline.md` edits will hit it again.

**What was deliberately NOT touched:** `gate-phase-entry.mjs`'s `IN_FLIGHT_MS` declaration and its exact source text, which #63-A2 pins. That pin passing is the same fact as this branch's **expected-survivor** mutation; softening it would have hidden precisely what #74 is about.

### Result

```
CI (ubuntu): All test suites passed.  45 suites, 4105 assertions, 0 failed.
             test-run-lifecycle.sh   31/0
             test-pipeline-status.sh 60/0
```

CI also confirms the `session-start.sh` portability design: the probe cell, which **asserts** rather than prints, reports `cutoff probe on this runner: 2026-09-01T10:29:39` — so the GNU `date -d` branch fired on Linux and the `updated_at` path was genuinely exercised there, not the mtime fallback.

Issue #140, opened 2026-09-03
An unbalanced quote defeats the blanket-staging gate in 59 bytes with no timeout, and the test suite documents the bypass as correct behaviour


nsmedia-io:
Found by SecOps during #132's Phase 2 round 2, confirmed by the orchestrator against the tree. Filed before #132 merges per the deferring-is-an-action rule. **This is not a timeout issue and is orthogonal to #116 and #132.**

## The bypass

An **unbalanced quote earlier in the same command** defeats the blanket-staging refusal. The scanner does not model heredocs, so a body's quote state leaks past the terminator and the staging verb on the far side is read as quoted text — which the gate deliberately allows.

**59 bytes. 91 ms. No timeout involved.** Every length and density bound in #116 and #132 is irrelevant to it.

| body | cmd bytes | decision | stdout | ms |
|---|---|---|---|---|
| 0 double quotes | 70 | deny | 471 | 177 |
| 3 double quotes | 60 | **none (ALLOWED)** | 0 | 88 |
| 4 double quotes | 63 | deny | 471 | 179 |
| 1 single quote | 59 | **none (ALLOWED)** | 0 | 97 |
| tracked `knowledge/issue-archive/40-verify-40.sh` as the body | 137,457 | **none (ALLOWED)** | 0 | 2,491 |
| the same body **+ one `"`** | 137,458 | deny | 471 | 2,726 |

The last two rows are a one-byte control: a single added quote character flips the verdict. The second-to-last row uses a **file already tracked in this repository**, so this is reachable with content the tree ships.

**Effective state is not raw parity.** `test-gate-phase-entry.sh` contains 5,183 double quotes and denies correctly. So a caller cannot predict which side of the bypass a given command lands on, and no length or density bound touches it.

## Why this was not found before: the suite documents it as correct behaviour

`plugins/pipeline/tests/test-pretooluse-gate-verdicts.sh:571-577` carries this comment, and it is the whole story:

> TRUNCATION MUST NOT LEAVE A QUOTE OPEN, and the first draft of this builder did. Cutting a generated body to a fixed length can end it inside a `"...`, and this scanner does not model heredocs -- so the terminator and the staging command on the far side of it were read as QUOTED TEXT and the row came back `none`. That would have been a fixture reporting an allow **for a gate that is correct**, which is the worst kind of red. The last two bytes are therefore spent balancing the quote counts.

[...]

Issue #138, opened 2026-09-03
hooks.json declares Stop timeout 600, which may equal the runtime's own default and therefore bound nothing tighter than absence


nsmedia-io:
Found by DevOps during #132's Phase 2 review and independently re-verified by the orchestrator against the shipped binary. Split out rather than folded into #132, whose scope is the PreToolUse bound only.

## The finding

`plugins/pipeline/hooks/hooks.json` declares four hook timeouts:

| hook | declared (seconds) |
|---|---|
| `pre-tool-use.sh` | 5 |
| `subagent-stop.sh` | 15 |
| `session-start.sh` | 20 |
| `stop.sh` | **600** |

`600` appears to be exactly the runtime's own default for the command-hook execution path. Verified directly from the shipped 2.1.85 binary, not relayed: the exec path reads `E=H.timeout?H.timeout*1000:EL`, and `EL=600000` (ms). So `timeout: 600` multiplies to 600000 — byte-identical to the fallback taken when the field is absent.

**A declared bound equal to the default bounds nothing tighter than absence.** It reads, to anyone auditing `hooks.json`, as a deliberate ceiling that somebody chose. It is not one. That is the same class #108 fixed when it corrected the other three from a milliseconds reading to seconds — a timeout field that looks decided and isn't.

## What is NOT established, and should be before this is closed

The binary contains **at least six distinct `.timeout*1000:<default>` sites with four different fallbacks** — `30000`, `60000`, `EL` (=600000), and `yof` (=600000). Which site serves which hook event was not determined.

This matters and is the first thing to settle:

- If the Stop hook's execution path is the `EL`/600000 site, the finding above stands: the declaration is a no-op.
- If it is the `30000` or `60000` site, the declaration is **not** a no-op — it is a 10x or 20x *widening* of the effective bound, which is a materially different (and more interesting) finding than a redundant one.

Do not close this on the assumption that 600 is universally the platform default. Establish which executor handles `Stop` first, from the binary, and record the witness.

## Why it was not fixed in #132

#132's scope is the PreToolUse bound, sized against a term (#116's linearisation) that no longer binds. `stop.sh` has no equivalent measurement behind it and no known fail-open incident. Folding an unmeasured second hook into a fix whose whole justification is a measurement would weaken both.

## What must be true of a correct answer

- The declared value must differ from whatever default the *Stop executor specifically* would apply, or the declaration must be removed rather than left as decoration. A field that cannot change behaviour should not look like it can.
[...]

Issue #137, opened 2026-09-03
_wordchunk's most-frequent-delimiter fallback is linear with a constant of 18, and #132's claim that it is the slowest shape measured is wrong by 3.2x


nsmedia-io:
Split out of #132 by the BA ruling at Phase 1, so the deferral is written down rather than claimed.

## The residual

`_wordchunk`'s fallback — the path taken when no single delimiter cuts an oversized whitespace-free word into pieces averaging under one window — picks the delimiter that occurred MOST. That bounds the product at `|_DELIMS| x length`. The bound is linear, but its constant is **18**.

Measured on a 64 KB whitespace-free word where every delimiter sits just under the acceptance threshold:

| implementation | time |
|---|---|
| `70e7d46` (pre-#116) | 4461 ms |
| first-present fallback | 3417 ms |
| most-frequent fallback (shipped) | 2251 ms |

## Correction to #132's characterisation of this residual

**#132's body states this is "the slowest per byte of any shape measured". That is wrong**, and the contradicting figures are in the same document: re-measured at Phase 1 it runs **0.034 ms/byte against quote-dense's 0.11** — 3.2x *cheaper*, not slower. The claim should not be carried forward into whatever work closes this.

It is also, as #132 correctly notes, the least likely shape to be written by accident.

## Why it was ruled out of #132

1. It is a **constant-factor optimisation of an already-linear path**, currently sitting at ~45% of the budget. #132 fixes a bound that no longer matches its binding term; this is tuning, and the two do not belong in one diff.
2. Touching `_wordchunk` re-opens **#116's 38,864-row differential burden** — the equivalence evidence that function's rewrite had to carry. That cost is not worth paying for a constant factor.

## Two figures to re-take when this opens

#132's body carries two numbers that did not reproduce at Phase 1 and should be re-measured here rather than trusted:

- `2251 ms` appears **twice for two different measurements** — the 350 KB one-unbroken-token row *and* this `_wordchunk` 64 KB fallback row. The 350 KB row re-took cleanly (2736 ms on a host running 1.22–1.32x slow, back-scaling to ~2200), so **this row is the suspect one**. It has the shape of a transcription collision.
- Separately, #132's "0.11 ms per KB" unit label is wrong by 1024x; the values are correct **per byte**. Read literally it implies 5 s buys 45,000 KB and there is no defect at all.

## What must be true of a correct answer

[...]

Issue #136, opened 2026-09-03
Move the staging gate's tokenizer into the resolver process, so the scan constant stops being a shell interpreter's (the expensive half of #132)


nsmedia-io:
Split out of #132 by the BA ruling at Phase 1, so the deferral is written down rather than claimed. #132 closes the *cheap* half (the declared bound is sized against a term that no longer binds after #116); this is the *expensive* half, and it is deliberately not a bug fix.

## The residual

After #116 the PreToolUse staging gate's scan is LINEAR in command length, but the constant is a shell interpreter's: roughly 16 us per statement in bash 3.2.57(1) in sh mode on darwin, ~8 us in dash. The scan turns once per structural character and the queue build once per word, so a large enough command still crosses whatever timeout `hooks.json` declares — and a PreToolUse hook killed at its declared timeout **emits nothing, which fails OPEN**.

Raising the declared timeout (#132) moves that crossing linearly and is a ratchet, not a treadmill. But a crossing exists at every finite T and **cannot be removed inside a POSIX sh implementation**, because the floor is one interpreter statement per structural character plus one per word.

## The candidate answer

The reject path is already decided in a handful of shell builtins — it forks nothing and starts no node process, which is the property #106 bought deliberately (a node cold start measured 66.68 ms against a 4.45 ms shell baseline, and paying that on every Bash tool call to answer "no" is a permanent tax).

Only the narrow *candidate* branch runs the tokenizer, and **that branch already starts a node process** for the run-owner resolver. Moving the tokenizer into that existing process would make its cost a rounding error at no extra process.

## Why this is not a free win, and why it is not a bug fix

What it costs is exactly the thing #106 refused on a stated security ground: **the caller's command text on a child's argv or stdin**. The argument for reopening is that stdin is not argv and is not readable by other local users, so this is a real option rather than a closed door.

That makes it a **new data flow, deliberately refused once already**. Three reasons it was ruled out of #132, in order of weight:

1. It re-opens a deliberate security refusal. A bug-fix scope does not admit that.
2. It needs a **SecOps reading, not a benchmark** — and a reading cannot be delivered on a deferred basis.
3. Its benefit is the *same kind* as raising the timeout: both move the crossing, neither removes it. So it is not what makes the cheap fix a false comfort.

This is an argument for doing it **later**, not instead.

## What must be true of a correct answer

- The verdict on every command must be unchanged from the shell-only path — the tokenizer moving processes is not licence to change what is denied.
[...]

PR #141, opened 2026-09-03
Raise the PreToolUse declared timeout out of the range the gate outruns, and stop the raise from widening four regression guards (#132)

### LINKED ISSUE #132: The PreToolUse staging gate's scan is now LINEAR in command length, but its constant is a shell interpreter's: a large enough command still outruns the declared 5 s and a killed hook fails open
## What #116 closed, and what it left

#116 took the PreToolUse staging gate's scan off the PRODUCT of structural density and command
length. The command is now split once by the shell's own field splitting into a queue of bounded
windows and the scan never holds more than one, so cost is LINEAR in total length and blind to how
the length is distributed. What it did NOT close is the CONSTANT, and the constant is a shell
interpreter's: about 16 us per STATEMENT in bash 3.2.57(1) in sh mode on darwin, about 8 us in
dash. The scan turns once per structural character and the queue build turns once per word, so a
large enough command still outruns the 5 s `hooks.json` declares -- and a PreToolUse hook killed at
its declared timeout emits nothing, which FAILS OPEN.

This is the residual #116 is closing WITH, recorded before that merge because a deferral only
counts once it is written down.

## The measurement

Same conditions for both sides, on darwin 25.5.0 / bash 3.2.57(1) in sh mode / node v24.19.0,
driving the shipped hook through `hooks.json`'s own command with a subagent payload and one
in-flight `4-review` record, against the declared 5 s. Old is `70e7d46`, new is #116's branch.
Every figure is the MINIMUM of three runs, taken at load average 4-7 on a host that had been idle
for several minutes: on a shared machine every sample carries load the gate did not cause, and load
only ever adds, so the minimum is the estimator. `struct` counts characters in the hook's own
`_STRUCT` set.

| shape | size | struct | old best | old under `timeout 5` | new best | new under `timeout 5` |
|---|---|---|---|---|---|---|
| quote-dense, `echo "a" ... ; git add -A` | 12.0 KB | 6002 | 4997 ms | **rc=124, ALLOWED** | 1535 ms | deny |
| the same | 14.0 KB | 7002 | 6067 ms | **rc=124, ALLOWED** | 1738 ms | deny |
| the same | 20.0 KB | 10002 | 10653 ms | **rc=124, ALLOWED** | 2216 ms | deny |
| JSON heredoc, 100 records | 19.2 KB | 3310 | 2963 ms | deny | 1323 ms | deny |
| the same, 150 | 29.0 KB | 4960 | 5514 ms | **rc=124, ALLOWED** | 2041 ms | deny |
| the same, 200 | 38.7 KB | 6610 | 8906 ms | **rc=124, ALLOWED** | 2697 ms | deny |
| prose heredoc, 20 000 words | 78.0 KB | 1675 | 3556 ms | deny | 1908 ms | deny |
| the same, 30 000 | 116.9 KB | 2508 | 5376 ms | **rc=124, ALLOWED** | 2949 ms | deny |
| `.pipeline/106/peer-review.json` + `git add -A` | 81.3 KB | 2519 | 3331 ms | deny (67% of budget) | 1420 ms | deny (28%) |
| the same file twice | 162.6 KB | 5029 | 12618 ms | **rc=124, ALLOWED** | 2007 ms | deny |
| `git commit -a -m "c <200000 y>"` | 200 KB | 2 | 3882 ms | deny | 909 ms | deny |
| the same at 350 KB | 350 KB | 2 | 10319 ms | **rc=124, ALLOWED** | 2251 ms | deny |

**Where it crosses now**, bracketed on the same host and the same day:

| shape | crossed before | crosses now | moved |
|---|---|---|---|
| quote-dense | ~12 KB | between 40 KB (4748 ms, deny) and 48 KB (6095 ms, ALLOWED) | 3.5x |
| ordinary JSON | ~29 KB | between 66 KB (4645 ms, deny) and 78 KB (5992 ms, ALLOWED) | 2.4x |
| quote-free prose | ~117 KB | between 195 KB (4670 ms, deny) and 234 KB (5658 ms, ALLOWED) | 1.8x |
| one giant token | ~230 KB | past 350 KB, which decides in 2251 ms | over 3x |

**And the shape of the curve changed, not only its height.** Six structural classes at one fixed
length, five lengths each, old and new measured BACK TO BACK on each fixture in the same run
(`120000+` is that measurement's own cap, not a figure; it was taken under load average 38-78, so
its absolutes are inflated against the table above and only its shape is being read here):

| body | 4 KB | 8 KB | 16 KB | 32 KB | 64 KB |
|---|---|---|---|---|---|
| **quote-dense** (2.0 B/struct) | 3717 / 1782 | 7734 / 2566 | 18942 / 5495 | 69423 / 9992 | 120000+ / 26036 |
| **minified JSON** (2.7 B/struct) | 2270 / 1956 | 6201 / 3160 | 17776 / 7802 | 58575 / 13827 | 120000+ / 24638 |
| **pretty JSON** (5.2 B/struct) | 1344 / 1030 | 3035 / 1676 | 7238 / 3506 | 19916 / 5899 | 57875 / 12978 |
| **source code** (5.7 B/struct) | 2276 / 1134 | 2116 / 2441 | 4572 / 4085 | 13255 / 7244 | 38348 / 13416 |
| **prose** (45 B/struct) | 1049 / 747 | 1001 / 931 | 1300 / 1357 | 2383 / 3030 | 5375 / 3897 |
| **one unbroken run** (O(1) struct) | 1893 / 549 | 6754 / 512 | 21567 / 1912 | 79750 / 1600 | 120000+ / 1901 |

Read DOWN a column and the old cost follows DENSITY; read ACROSS a row and it follows the SQUARE of
length. Neither is true of the new one. At 64 KB the spread across that whole density range fell
from over 22x (three cells past the cap, so the true figure is larger) to 13.7x.

## What remains, stated as the thing a reader should triage on

**The cost is linear and the constant is large.** On an idle host the new scan decides a
quote-dense command at about 0.11 ms per KB, ordinary JSON at 0.07, prose at 0.025 and one unbroken
token at 0.006. A command large enough still crosses 5 s and is then ALLOWED -- from about 42 KB of
the densest shape, 70 KB of ordinary JSON and 205 KB of prose, against 12, 29 and 117 KB before.
The crossing moved out by 1.8x to over 3x depending on the shape; it did not go away, and it cannot
go away inside a POSIX sh implementation, because the floor is one interpreter statement per
structural character plus one per word.

Two things follow, and they are different sizes of answer:

1. **The cheap one: the bound is a declaration, not a law.** `hooks.json` declares `timeout: 5`. The
   platform's own default is 600 s. Raising the PreToolUse timeout moves the fail-open point
   linearly for free, at the cost of a slower worst case on a Bash call that is already pathological.
   Nothing about that is a redesign, and it is the first thing to consider.
2. **The expensive one: the language.** The two-stage split (#106) exists because a node cold start
   measured 66.68 ms against a 4.45 ms shell baseline and paying it on every Bash tool call to
   answer "no" is a permanent tax. But the reject path is already decided in a handful of shell
   builtins; only the narrow candidate branch runs the tokenizer. Moving THAT branch into the
   resolver that already starts a node process would make the tokenizer's cost a rounding error and
   would cost no extra process, because the process is started anyway. What it would cost is the
   thing #106 deliberately refused: the caller's command text on a child's argv or stdin. Stdin is
   not argv and is not readable by other local users, so this is a real option rather than a closed
   door -- it needs a SecOps reading, not a benchmark.

**A second, narrower residual, with its own measurement.** `_wordchunk`'s fallback -- the path taken
when no single delimiter cuts an oversized whitespace-free word into pieces averaging under one
window -- now picks the delimiter that occurred MOST, which bounds the product at
`|_DELIMS| x length`. That bound is linear but its constant is 18. Measured on a 64 KB
whitespace-free word where every delimiter sits just under the acceptance threshold: 4461 ms at
`70e7d46`, 3417 ms with a first-present fallback, 2251 ms with the most-frequent one. It is the
slowest per byte of any shape measured, and it is also the least likely to be written by accident.

## Two controls in the shipped code that no live input can defeat today

Recorded because a mutation battery run against #116's differential found them, and a surviving
mutation that is a THEOREM should be declared as one rather than left looking like a coverage gap.

* **`_wordchunk`'s leading-newline restore.** Deleting it moves no verdict over 38 864 corpus rows,
  because every queue entry begins with the space level 1 prepends unconditionally, so `_wc_w` can
  never open with a newline. It is a ratchet on `_wordchunk`'s own contract; a level-1 build that
  stopped prepending would need it, and would otherwise merge two invocations silently.
* **The two-character window guard in `_scan_go`.** Weakening `??*` to `?*` moves no verdict, for a
  reason worth writing down: window edges fall at word boundaries (level 1) and immediately before a
  delimiter (level 2), and the only pair whose split is not inert is `>&`, which cannot sit at either
  kind of edge inside a git invocation -- a level-2 edge exists only inside a word of 1024 bytes or
  more, and that word's own `&` characters have already ended the invocation before the `>&` is
  reached. Also a ratchet.

## Direction and reachability

A LIVE INPUT still reaches it -- no privilege, no future edit -- and it is still a SILENCE rather
than a falsehood: past the crossing the gate goes quiet where it should refuse, which is the
documented fail-open direction for every other tooling gap in this hook. It refuses no correct work
at any size. What changed is the size, by about 3x on every class measured, and the SHAPE of the
curve, which no longer multiplies two terms together.

## How to re-take it

The generators, the paired driver and the 38 864-row differential corpus used above are described in
#116 and reproduced by the same commands; `plugins/pipeline/tests/test-pretooluse-gate-verdicts.sh`
carries the DENSITY AXIS and DENSITY LINEARITY blocks, which assert the property in-tree at 8 KB and
record the observed curve on whatever host runs them.

nsmedia-io:
Closes #132

**DRAFT DELIBERATELY.** Two rows are red at this HEAD and neither is Dev's to clear; both are named below. Merging before they clear ships a red main, because `test-issue17-integration.sh` AC41(c) asserts `run.sh` exits 0 in a fresh clone and CI runs it on every push to main.

## What this changes

The PreToolUse Bash gate declared a 5-second `timeout`. This repository's own largest tracked artifact -- `knowledge/issue-archive/106.json` written as a heredoc body and staged with `git add -A`, **461795 command bytes at 23.92 B/struct** -- decides `deny` in **7004 ms** min-of-3 on darwin 25.5.0 at load 2.19, which is **13128 ms** once the load (1.42) and cross-host (1.32) spreads the spec records are applied. **A PreToolUse hook killed at its declared timeout emits nothing, and a hook that emits nothing FAILS OPEN**, so past that length the blanket staging the gate exists to refuse was allowed.

The declaration is **30 s**. The value is bound by AC2's inequality against the WORST of six recorded observations of that fixture (12203 ms at load ~12, giving 22873 ms adjusted, which demands at least 23 s). 15 and 20 are both refused by that arithmetic; 30 is the ceiling of the existing `0 < t <= 30` range assertion and 20x below the platform's 600 s default. Re-verified under load 12.75 during the full local pass: 8481 ms min-of-3, 15896 ms adjusted, 1.89x headroom.

Alongside it, so the raise cannot be a one-integer edit that quietly widens things:

- **The four timing probes' regression budget is now absolute** (`REGRESSION_BUDGET_MS=5000`, frozen at the value those guards already carried) and separate from the declared timeout, which survives as a distinct BYPASS assertion in each block. Both bounds are asserted with distinguishable failure messages.
- **The declared value is pinned exactly**, fail-open tail included. It was the only one of the four `hooks.json` entries checked by a range: the declaration suite reported 68 passed / 0 failed at 5, 15, 19 AND 30 alike, on the security-critical entry.
- **The sizing corpus is a rule over the tree at check time**, not a list of filenames: content-only, extension-open, floor 2000 bytes, top-5 by raw length unioned with top-5 by raw B/struct plus two sentinels from outside both, every row driven with its ALLOW control. It enumerated 163 files and drove 12. Phase 5 writes one `knowledge/issue-archive/<n>.json` per issue forever, so the published count is a tripwire: the suite reddens when the corpus outgrows the bound instead of the gate falling open.
- **README item 27 cost (4) is re-derived**, with both density ends measured against the new bound, the cells that remain uncovered named with their milliseconds and their density, the ALLOW-path stall published (7120 ms against the denied arm's 7004 on the identical body), the growth direction stated, and #116's superseded pair re-taken rather than left standing. Upgrading gains item 30 and both count words move to thirty.
- **A knowledge-store checker** reconciles the store's stated timeout against `hooks.json`, anchored on the SUBJECT rather than on a phrase, and reports what it compared including when it compared nothing.

## Two defects this work surfaced

**Linux silently truncated every large fixture command, and only on the host CI runs.** Linux caps a single argv string at MAX_ARG_STRLEN (131072 bytes) independently of ARG_MAX; macOS does not. `gate_payload` passed the whole command as one argv string, so on ubuntu-latest the 461795-byte floor fixture failed the exec, came back as an EMPTY payload, and the gate answered `none` in **110 ms with zero bytes of stdout** -- a fixture recording an ALLOW for a gate that was working. Fixed here; the two input paths are asserted byte-identical.

**A naive quote-parity pad broke the thing it guarded.** Counting raw quotes is on the wrong side of the transformation the scanner applies: `106.json` carries an odd number of raw double quotes and is refused correctly, and appending a balancing one flipped that `deny` to `none`. The pad is searched by outcome now, and the rows that need one are recorded -- on this tree that is `README.md`, a live instance of **#140** on tracked content.

## The two red rows

1. **Three `[PHASE-5]` rows** in `test-pretooluse-timeout-knowledge.sh` (plus four consequential rows in `test-issue17-integration.sh`). `knowledge/living-context/architecture--pretooluse-tracked-write-gate.json` still says `5-SECOND` and still frames the declaration as comfortably sized. It is Librarian-owned; Dev is read-only on it. Every discriminating mutation in AC17/AC18 runs against a materialized COPY, so correcting the record turns these green with the suite byte-identical. **The Librarian commit must land on this branch before merge, and after this README rewrite**, per BA's round-3 ruling.
2. **`AC11 SIZE CEILING`**, red on ubuntu-latest only. QA's fail-open fixture fits its body from two points at 26 KB and 106 KB and extrapolates to tens of megabytes; on dash that fit overshoots 15.6x (aimed at 45000 ms, measured 700683 ms at 23159538 bytes), busting the 6 MB ceiling. No value in the admissible window \[15,30\] satisfies that ceiling on that host, so it is not a value choice. Raised to QA rather than edited; details and a recommendation are in `impl-report.json`.

## Cost, disclosed rather than hidden

The CI `suite` job went from roughly 4 minutes to **35 min 6 s** (run 33747342504). About 23 of those minutes are the single AC11 call above, paid twice because `test-issue17-integration.sh` runs `run.sh` inside a clone. Correcting that fit removes most of it. The sizing-corpus block this PR adds is the small term: +64 s on darwin, +3 s on ubuntu-latest.

## Residuals filed and deliberately out of scope

**#136, #137, #138 and #140.** #140 in particular is live and needs no timeout at all: an unbalanced quote earlier in the same command defeats the blanket-staging refusal outright.


nsmedia-io:
## Phase 4 Peer Review (architectural tier panel, round 2)

| Agent | Verdict | Notes |
|---|---|---|
| BA | APPROVE | Spec intent honored; CI-fix commits judged in-scope (forced by CI grading against a moving target, not drift) |
| DBA | APPROVE | No data-layer surface; AC4 population rule live-reverified against actual HEAD |
| DevOps | APPROVE_WITH_NOTES | ref-pin fix verified correct on both triggers; 2 nits, both non-actionable (see below) |
| SecOps | APPROVE_WITH_NOTES | ref-pin fix confirmed currently safe (no pull_request_target, no secrets, no elevated permissions); 1 low note on the general reuse-risk pattern |
| Dev | APPROVE | |
| QA | APPROVE_WITH_NOTES | Binding test verdict. 1 low note, host-load flakiness independently re-verified clean against real CI green |

**Final verdict:** APPROVE_WITH_NOTES

All three notes checked and found non-actionable against this diff:
- DevOps's ten-file replicated-span digest mismatch claim does **not** reproduce — verified independently against both `origin/main` and this HEAD using the validated extraction method; all ten files agree at the reference digest `847cd28217115c41dc8628cb8e35a4f9162c5bfe`. This finding hit the exact trailing-newline false-alarm case the block's own documentation warns readers about.
- DevOps + QA's wall-clock assertion flakiness (`test-pretooluse-gate-declaration.sh`, `test-pretooluse-timeout-bound.sh`) is attributed to this session's host CPU contention (load averages 23–68 observed from unrelated concurrent work), and both independently confirmed CI itself is green at this exact commit.
- SecOps confirmed the CI checkout-ref pin (`714e643`) is currently safe and is a review-integrity improvement (closes the moving-target problem where CI graded a PR against a shifting `main`), flagging only the general risk shape for future awareness.

CI reconfirmed green at merge time: `manifests` and `tests` both `success` at `714e643ced874e3901cc20bdee875e8e16aeb79f`.

